PR #8, opened 2016-02-11
API updates

### LINKED ISSUE #2: Support building UFOs
README says, 

> This library provides a wrapper for several other Python libraries which together compile fonts from various sources (.glyphs, .ufo) into binaries (.otf, .ttf).

but it seems to only accept .glyphs files atm,

https://github.com/googlei18n/fontmake/blob/master/Lib/fontmake/__main__.py#L22

In reading https://github.com/googlei18n/fontmake/blob/master/Lib/fontmake/font_project.py then I suppose there needs to be a default assumption that an input UFO is a master or an instance, and an arg to declare this I propose that assuming UFOs are masters makes more sense.

### LINKED ISSUE #5: Add cmdline tool
Using setup.py, you can make setup create a cmdline tool called fontmake that will run the module.  See:

https://github.com/trufont/trufont/blob/master/setup.py#L41

but use `console_scripts` instead of `gui_scripts`.

### LINKED ISSUE #6: Add option for using makeotf instead of feaLib
For now, I’d feel better if it was possible to choose between feaLib and makeotf for compiling OpenType features. Once feaLib is known to work well, we can remove the option.

jamesgk:
Adds a few things:
- command line tool "fontmake" (to close #5)
- option to build from UFO sources (to close #2)
- option to build using the AFDKO to compile features (to close #6)

cc @brawer @behdad @davelab6

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 
+import glob
 import os
 import plistlib
 import re
@@ -25,6 +26,19 @@ from glyphs2ufo.glyphslib import build_masters, build_instances
 from ufo2ft import compileOTF, compileTTF
 from ufo2ft.kernFeatureWriter import KernFeatureWriter
 
+OpenUfo = None
+try:
+    from defcon.objects import Font
+    OpenUfo = Font
+except ImportError:
+    pass
+try:
+    from robofab.world import OpenFont
+    OpenUfo = OpenFont
+except ImportError:
+    pass
+if OpenUfo is None:
+    raise ImportError("Couldn't import from defcon or robofab.")
 
 class FontProject:
     """Provides methods for building fonts."""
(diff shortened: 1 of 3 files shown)


dougfelt:
LGTM modulo a few questions

davelab6:
Awesome, thanks!

Lib/fontmake/font_project.py
```diff
@@ -197,3 +214,49 @@ class GlyphsKernWriter(KernFeatureWriter):
...
+        fd, outline_path = tempfile.mkstemp()
+        os.close(fd)
+        self.outline.save(outline_path)
+
+        fd, feasrc_path = tempfile.mkstemp()
+        os.close(fd)
+
+        fd, fea_path = tempfile.mkstemp()
+        with open(fea_path, "w") as feafile:
+            feafile.write(self.features)
```
adrientetar: os.write(fd, self.features)

PR #32, opened 2016-02-18
Do compound transform in decomp, using fontTools

jamesgk:
This is in response to
https://github.com/googlei18n/fontmake/issues/17#issuecomment-185555613


behdad:
Is this still relevant?

My suggestion was to merge scale and offset into one Transform.  Feel free to leave code as is for now.

PR #36, opened 2016-02-24
Minor API change

jamesgk:
I'm thinking this is not very important, but I had it locally and decided to push it. I could take it or leave it as far as merging.


dougfelt:
LGTM

PR #39, opened 2016-03-08
Don't use special kern writer for Glyphs

jamesgk:
glyphs2ufo now outputs kerning rules with ufo3-style group names (as of https://github.com/googlei18n/glyphs2ufo/commit/d2b1a23b3af16e3a04bb52d038f5cd44654514f0), so there's no need for a special kern writer class.


dougfelt:
LGTM

PR #52, opened 2016-03-25
MutatorMath input

jamesgk:
cc @brawer

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -25,6 +25,7 @@ from booleanOperations import BooleanOperationManager
 from cu2qu.rf import fonts_to_quadratic
 from fontTools import subset
 from glyphs2ufo.glyphslib import build_masters, build_instances
+from mutatorMath.ufo import build as build_designspace
 from ufo2ft import compileOTF, compileTTF
 from ufo2ft.makeotfParts import FeatureOTFCompiler
 from ufo2ft.kernFeatureWriter import KernFeatureWriter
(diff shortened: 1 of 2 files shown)


dougfelt:
LGTM

Lib/fontmake/font_project.py
```diff
@@ -25,19 +25,12 @@
 from cu2qu.rf import fonts_to_quadratic
 from fontTools import subset
 from glyphs2ufo.glyphslib import build_masters, build_instances
+from mutatorMath.ufo import build as build_designspace
+from robofab.world import OpenFont
```
anthrotype: hmm. why do you [drop the robofab dependency in ufoLib](https://github.com/googlei18n/ufo2ft/pull/31) while at the same time you require it in fontmake?

anthrotype: sorry I just noticed the commit message:

> For the time being cu2qu only works with robofab, so don't try using defcon.

well, we shall fix this in cu2qu then. Once the pens are merged in cu2qu, defcon should work.

adrientetar: > Once the pens are merged in cu2qu, defcon should work.

Yeah but the FontCollection stuff that generates compatible fonts in parallel won't work with a basic pen, will it?

anthrotype: no, not yet. If have some ideas as to how to adapt the pen protocol to the *Collection classes in `cu2qu.rf` module, but don't have anything to show yet. Suggestions are welcome.

PR #55, opened 2016-04-01
Use defcon

jamesgk:
This is mostly to use defcon throughout the toolchain, but also includes a little cleanup.


dougfelt:
LGTM

Lib/fontmake/font_project.py
```diff
@@ -75,32 +77,26 @@ def remove_overlaps(self, ufo):
         for glyph in ufo:
             self.decompose_glyph(ufo, glyph)
             manager = BooleanOperationManager()
-            contours = glyph.contours
+            contours = [c for c in glyph]
```
behdad: Would `list(glyph)` do this?

behdad:
lgtm.

We should also start thinking about putting together a test suite for fontmake.

PR #58, opened 2016-04-07
Separate compatible output directory

jamesgk:


Lib/fontmake/font_project.py
```diff
@@ -236,7 +238,9 @@ def _font_name(self, ufo):
     def _output_dir(self, ext, is_instance=False):
         """Generate an output directory."""
 
-        dir_prefix = 'instance_' if is_instance else 'master_'
+        dir_prefix = os.path.join(
+            'out_compatible' if self.compatible else 'out',
```
behdad: humm.  I was expecting something like master_ttf_interpolatable or something.

Lib/fontmake/font_project.py
```diff
@@ -185,16 +188,15 @@ def run_from_designspace(self, designspace_path, **kwargs):
...
+            self, ufos, remove_overlaps=True, mti_source=None, **kwargs):
         """Run toolchain from UFO sources to OpenType binaries."""
 
         if isinstance(ufos, str):
             ufos = glob.glob(ufos)
         if isinstance(ufos[0], str):
             ufos = [Font(ufo) for ufo in ufos]
 
-        if remove_overlaps and not compatible:
+        if remove_overlaps and not self.compatible:
```
behdad: Can you also make sure `remove_overlaps` is turned off if compatible is turned on?  Or fail if they conflict?  Right now `remove_overlaps` is always on, right?  For GX / interpolation, we need it turned off.

Maybe instead of calling it `compatible`, we should call it something better.  But we can do that later when we figure out what to call these "GX" fonts.

behdad:
lgtm with comments.  Thanks.

Lib/fontmake/font_project.py
jamesgk: That could also work, though my thinking was that twelve base output directories `(master/instance)_(ufo/otf/ttf)[_interpolatable]` gets a bit messy, so better to put everything in either an out/ or out_interpolatable/ directory. But it's subjective.

Lib/fontmake/font_project.py
jamesgk: This check does prevent overlap removal when compatibility is turned on. But we can make it fail instead of silently doing this.

jamesgk: Actually, there's not even a command line argument to turn overlap removal on/off. Maybe the simplest thing to do is just only check `self.compatible` here and remove the remove_overlap argument altogether, assuming it's always wanted unless compatibility is required.

Lib/fontmake/font_project.py
behdad: For ufo and otf, interpolatable doesn't make sense.  They are always interpolatable.  It only matters for TTF.  And for instances it doesn't make any sense either.

It's one configuration that I need: "TrueType outlines, interpolatable, overlaps-retained, masters".  Just add a config for that particular case?

Lib/fontmake/font_project.py
behdad: Yeah, I suggest you make '--compatible' produce the exact thing we need for "GX" fonts.  There's no other usecase for it really.

jamesgk:
PTAL? Sorry this PR just got a lot bigger. I decided to take this opportunity to expand the API for different types of output (TTF, OTF, compatible).

behdad:
lgtm.  The last commit, I think it means ttf-interpolatable and --instances can't run together?  The error message could be less cryptic.

PR #66, opened 2016-04-21
Use Timer from fontTools for profiling

jamesgk:


jamesgk:
Hmm. This doesn't seem to work when mutatorMath is used. That is, when interpolating glyphs source or when given a designspace source.

anthrotype:
Because MutatorMath library is using the logging module in the wrong way. I'll send a pull request to fix that.

anthrotype:
James, the problem was only indirectly related to MutatorMath. The main problem was that you forgot to configure logging; you were relying on the "last resort" handler: i.e. the one which outputs to stderr when the user does not explicitly set up any handler for the root logger (and only if log messages level is >= WARNING). On the other hand, MutatorMath is calling basicConfig (whereas it shouldn't, TL;DR see below); hence, the last-resort stderr handler is not used when fontmake uses mutatorMath, and since you did not configure it, you see no messages from your timer logger.

You should fix this by simply calling `logging.basicConfig` yourself from within fontmake's `main()` function.
I suggest that you decrease the level of the timer's logger to "DEBUG", as "WARNING" is too high for this sort of information.
Here's the patch that solves this particular fontmake issue.

```diff
diff --git a/Lib/fontmake/__main__.py b/Lib/fontmake/__main__.py
index b5ec473..0d344da 100644
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -15,6 +15,7 @@
 
 from argparse import ArgumentParser
 from fontmake.font_project import FontProject
+import logging
 
 
 def main():
@@ -31,7 +32,10 @@ def main():
     parser.add_argument('--timing', action='store_true')
     args = vars(parser.parse_args())
 
-    project = FontProject(timing=args.pop('timing'))
+    timing = args.pop('timing')
+    if timing:
+        logging.basicConfig(level="DEBUG", format="%(message)s")
+    project = FontProject(timing=timing)
 
     glyphs_path = args.pop('glyphs_path')
     ufo_paths = args.pop('ufo_paths')
diff --git a/Lib/fontmake/font_project.py b/Lib/fontmake/font_project.py
index 57f43b4..c5c5ce2 100644
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -36,7 +36,7 @@ from mutatorMath.ufo.document import DesignSpaceDocumentReader
 from ufo2ft import compileOTF, compileTTF
 from ufo2ft.makeotfParts import FeatureOTFCompiler
 
-timer = Timer(logging.getLogger('fontmake'), level=logging.WARNING)
+timer = Timer(logging.getLogger('fontmake'), level=logging.DEBUG)
 
 
 class FontProject:
```

Now, the way MutatorMath uses logging is rather uncommon for a library.
It is usually the user of the a library who's responsible to configure any loggers (e.g. attaching a file or stream handler, with optionally some format strings and verbosity level, etc.), not the library itself.

The logging configuration should be done inside a script's main() function. However, MutatorMath was calling the `basicConfig` function from the constructor of `DesignSpaceDocumentReader`, which fontmake uses. The problem is that configures the handler for the "root" logger, which is the parent of all the loggers; so it has an effect not only on MutatorMath's logger, but on any other library which happen to use `logging` module. I'm going to send a PR for this specific issue to the upstream MutatorMath repo.






`logging.basicConfig`

anthrotype:
On second thought, it's better you don't use basicConfig in fontmake.main() as I suggested early on, because that is going to turn "DEBUG" messages from all loggers (as that configures the root logger). Instead, you can use fontTools.misc.loggingTools.configLogger function, which works like logging.basicConfig but lets you specify a specific logger instance to configure. This way you can quickly add a handler only for the logger used in the Timer, and avoid DEBUG messages coming from other libraries (eg fontTools). I hope this makes sense.

jamesgk:
Thank you Cosimo! I am indeed very unfamiliar with the logging module. How does this look?

anthrotype:
👍

anthrotype:
this is ok for now. You just configured the 'fontmake' logger, which for now only output debug (timing) messages.

Later maybe you may wish to turn all `print()` calls into equivalent `log.info`, etc. calls, and allow the user to set the verbosity of the logging from fontmake.main().

As more and more libraries add logging support, an application such as fontmake will have to somehow configure the root logger, or set special handlers/formatters for each library's logger (I can help you with that).

PR #69, opened 2016-04-28
Rename glyphs

jamesgk:
This adds a glyph rename operation as the last compilation step in fontmake. I'm actually not sure if this is the best place for the production name generation logic. Maybe it should be in glyphsLib, and every glyph in the UFOs should have production name entries? Maybe the subset/rename stuff should all be in ufo2ft?

Related to https://github.com/googlei18n/glyphsLib/issues/12
cc @behdad @schriftgestalt

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -219,12 +219,9 @@ class FontProject:
         if ufo.lib.get(GLYPHS_PREFIX + "Don't use Production Names"):
             return
 
-        rename_map = {}
-        for glyph in ufo:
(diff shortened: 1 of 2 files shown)


khaledhosny:
I think ufo2ft would be a better place to do the renaming, since this shouldn’t be Glyphs-specific. I’m building a font that uses an Adobe font where production names are stored in `GlyphOrderAndAliasDB` which I want to parse and set `public.postscriotName` based on it.

jamesgk:
Thanks for the feedback @khaledhosny. PR opened on ufo2ft.

PR #70, opened 2016-04-29
Account for renaming in subset step

jamesgk:
This uses the new glyph names from the output binary to define a
glyph subset. It also uses a period as the script suffix delimiter
since that's what ufo2ft expects.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -182,10 +182,7 @@ class FontProject:
     def subset_otf_from_ufo(self, otf_path, ufo):
         """Subset a font using export flags set by glyphsLib."""
 
-        font_lib_prefix = GLYPHS_PREFIX
-        glyph_lib_prefix = font_lib_prefix + 'Glyphs.'
-
-        keep_glyphs = set(ufo.lib.get(font_lib_prefix + 'Keep Glyphs', []))
+        keep_glyphs = set(ufo.lib.get(GLYPHS_PREFIX + 'Keep Glyphs', []))
 
         include = []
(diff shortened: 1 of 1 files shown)


Lib/fontmake/font_project.py
```diff
@@ -175,19 +182,20 @@ def save_otfs(
     def subset_otf_from_ufo(self, otf_path, ufo):
         """Subset a font using export flags set by glyphsLib."""
 
-        font_lib_prefix = 'com.schriftgestaltung.'
+        font_lib_prefix = GLYPHS_PREFIX
```
dougfelt: Shouldn't you just replace the two uses of font_lib_prefix with GLYPHS_PREFIX?

jamesgk: Maybe. Though this way the prefixes for the font lib and glyph lib are more explicitly distinguished. We could do without the `glyph_lib_prefix` variable also?

dougfelt: I guess I'm not sure of the semantic distinctions you're trying to maintain.  If it is just a way to avoid typos, then I guess I would just have a single global define for each prefix I want to use in more than one place, and you just concatenate something onto the prefix.  If there's semantic distinctions that you want to be clear about even when the prefixes end up being the same in particular cases, then I'm unsure what to do.

jamesgk: Yea maybe it's simpler just to go with the former strategy.

jamesgk:
PTAL

dougfelt:
LGTM

PR #76, opened 2016-05-18
Use local sources for CI tests

jamesgk:


requirements.txt
```diff
@@ -1,18 +1,18 @@
 # for MutatorMath
-git+https://github.com/typesupply/fontMath.git@ufo3
-git+https://github.com/unified-font-object/ufoLib.git
+git+https://github.com/typesupply/fontMath.git@e96af7a672aa8173139952cf78405b6c43fd5353
```
anthrotype: I approve 👍 

But hopefully we shall soon have proper version numbers instead of commit hashes.

PR #93, opened 2016-06-20
Add fontdiff step to integration test

jamesgk:
This adds some simple rendering tests to the integration test. Ideally
we could add some more complex text to render (for example UDHR text),
but there's actually some kerning differences between the golden data
and fontmake output so that will have to wait (or, we could just add
test cases we know will fail and then try to fix them very soon).

cc @brawer


brawer:
👍

PR #96, opened 2016-06-28
Add option to keep overlaps

moyogo:
Some font source may already have removed overlap, or one may not want to remove overlaps. booleanOperations cannot yet remove overlap for quadratic curves: https://github.com/typemytype/booleanOperations/issues/26.


anthrotype:
👍

typemytype:
or help making it work :) 
I guess that would be the better option...
(I had no time so far)

PR #98, opened 2016-06-30
Add family name parameter

jamesgk:


jamesgk:
Oh, this needs an updated requirements.txt once the corresponding PR in glyphsLib is merged before it will pass the CI test.

PR #101, opened 2016-07-05
Add option for production names

moyogo:
Add option to use production names or not as alternative to default which uses production name unless disabled with glyph.lib's com.schriftgestaltung.Don't use Production Names

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -167,7 +167,8 @@ class FontProject:
     @timer()
     def save_otfs(
             self, ufos, ttf=False, interpolatable=False, mti_paths=None,
-            is_instance=False, use_afdko=False, subset=True):
+            is_instance=False, use_afdko=False, subset=True,
+            use_production_names=None):
         """Write OpenType binaries."""
 
         ext = 'ttf' if ttf else 'otf'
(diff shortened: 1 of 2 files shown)


anthrotype:
👍

moyogo:
I piled up minor fixes on top.

Lib/fontmake/font_project.py
```diff
@@ -168,7 +168,7 @@ def build_interpolatable_ttfs(self, ufos, **kwargs):
     def save_otfs(
             self, ufos, ttf=False, interpolatable=False, mti_paths=None,
             is_instance=False, use_afdko=False, subset=True,
-            use_production_names=None):
+            use_production_names=None, **kwargs):
```
jamesgk: Instead of allowing any keyword argument here, I think we should exclude `family_name` from args in \__main__.py similar to how we exclude `interpolate`, except between checking glyphs_path and designspace_path.

Maybe make a function which performs the check given an excluded argument name (or better, list of argument names) and allowed source names (to use in the error message), and call that function once with `['family_name'], 'Glyphs'` and then with `['interpolate'], 'Glyphs or MutatorMath'`.

moyogo: @jamesgk Is this better?

jamesgk: Perfect, thanks

PR #102, opened 2016-07-06
Add option to keep direction since default reverses contour direction

moyogo:
This can be used along with --keep-overlaps, in case the source already has the right contours.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -31,6 +31,9 @@ def main():
     parser.add_argument('--family-name', help='Family name to use for masters,'
                         ' and to filter output instances by')
     parser.add_argument('--use-afdko', action='store_true')
+    parser.add_argument('--keep-direction', dest="reverse_direction",
+                        action='store_false', help='Do not reverse contour '
+                        'direction when output is ttf or ttf-interpolatable')
     parser.add_argument('--keep-overlaps', dest="remove_overlaps",
                         action='store_false',
                         help='Do not remove any overlap.')
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -128,13 +128,15 @@ class FontProject:
             component.draw(pen)
 
     @timer()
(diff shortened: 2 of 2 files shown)


moyogo:
Travis CI is failing because this PR requires https://github.com/googlei18n/fontmake/pull/101/commits/46f06ea496c808eebceef974723cbbb7acfe8f23

Lib/fontmake/__main__.py
```diff
@@ -31,6 +31,9 @@ def main():
     parser.add_argument('--family-name', help='Family name to use for masters,'
                         'and to filter output instances by')
     parser.add_argument('--use-afdko', action='store_true')
+    parser.add_argument('--keep-direction', dest="reverse_direction",
+                        action='store_false', help='Do not reverse contour '
+                        'direction')
```
jamesgk: Maybe we should clarify that this only applies to contours in truetype output.

anthrotype: I think we need a way to skip cu2qu altogether when one provides input UFOs which already contain quadratic curves. There's no point in passing through cu2qu if one already knows the input sources are already in quadratic. We were thinking about fontmake looping through the glyphs and breaking as soon as it finds a 'curve' segment if the output format is ttf, or a 'qcurve' segment if the output is otf (though converting 'qcurve' to 'curve' is not supported yet).

anthrotype: or maybe make sure cu2qu does not modify the input fonts unless it finds a cubic curve segment.
The GetSegmentsPen that gathers the contour data could be modified to also record whether a glyph contains any curve or qcurve. This way one could skip the cu2qu conversion if it's not necessary.
https://github.com/googlei18n/cu2qu/blob/master/Lib/cu2qu/ufo.py#L97

jamesgk: We could do away with the reverse_direction parameter in cu2qu and do winding direction testing there, only reversing if necessary for truetype.

PR #104, opened 2016-07-13
Add option to run ttfautohint

jamesgk:


dougfelt:
LGTM

davelab6:
I have a trivia question :) Why in https://github.com/googlei18n/fontmake/blob/master/Lib/fontmake/ttfautohint.py#L34-L42 are the options underscore separated, instead of dash separated, then requiring the conversion of underscores to dashes?

jamesgk:
That only applies to keyword arguments passed into the function in python, not to a string of command line arguments (where you should be able to use underscores or dashes). There's no way to pass keywords with dashes in python.

anthrotype:
Because the hypen is not a valid character for Python variables.

jamesgk:
Sorry, I was mistaken there, you can't use underscores when passing in a string of arguments. So kwargs have to have underscores, and a string of args has to have dashes.

davelab6:
Got it! Thanks! :)

PR #108, opened 2016-07-21
Add option to treat masters as instances

jamesgk:
This should allow building masters and instances with `-i -mi`. All outputs will be put into instance output directories. Resolves https://github.com/googlei18n/fontmake/issues/107
r? @khaledhosny


khaledhosny:
Works just fine, thanks.

PR #111, opened 2016-07-26
avoid converting the curves twice with latest ufo2ft

anthrotype:
The current ufo2ft has a new `convertCubics=True` argument  which conflicts with fontmake, as the latter already calls `cu2qu.font[s]_to_quadratic` before passing the font to ufo2ft.
So fontmake needs to set that to False.


anthrotype:
ping @jamesgk

anthrotype:
Thanks!

PR #112, opened 2016-07-26
do not always run the fonttools subsetter

anthrotype:
`save_otfs` method has a `subset` keyword argument which defaults to `True`, without a command line option to disable it.
The fonttools Subsetter is called on the generated font, even if the input UFO does not have any `com.schriftgestaltung.Keep Glyphs` key in the `lib.plist`, nor any of the .glif libs have `com.schriftgestaltung.Glyphs.Export` set to False.

This PR set the default value of 'subset' argument to `None`, and adds two command line options `--subset` and `--no-subset`.

`None` means run the subsetter only if the "Keep Glyphs" list is present in the lib.plist.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -175,7 +175,7 @@ class FontProject:
     @timer()
     def save_otfs(
             self, ufos, ttf=False, interpolatable=False, mti_paths=None,
-            is_instance=False, use_afdko=False, autohint=None, subset=True,
+            is_instance=False, use_afdko=False, autohint=None, subset=None,
             use_production_names=None):
         """Write OpenType binaries."""
 
@@ -199,6 +199,8 @@ class FontProject:
                 convertCubics=False)
             otf.save(otf_path)
 
+            if subset is None:
+                subset = (GLYPHS_PREFIX + 'Keep Glyphs') in ufo.lib
             if subset:
                 self.subset_otf_from_ufo(otf_path, ufo)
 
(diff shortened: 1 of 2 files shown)


anthrotype:
BTW, do we actually need to use the subsetter at all here?
ufo2ft is already creating a new font from scratch, it could well generate only the desired subset.

The immediate reason I want to disable the subsetter in fontmake is that it drops the tables it can't subset (e.g. TSI tables from VTT). This could be "fixed" in fonttools by adding another option to `--pass-through` unknown tables, but that is not the point here.

khaledhosny:
I think whether we need the subsetter or not depends on Glyphs behavior, is the specified subset the exact (final) list of the glyphs to keep or just the base one.

anthrotype:
In any case, I just want a `--no-subset` option to avoid that.

jamesgk:
> BTW, do we actually need to use the subsetter at all here?
> ufo2ft is already creating a new font from scratch, it could well generate only the desired subset.

I think the reason I didn't want to do this is that it gets overly complicated when you consider opentype features.

Lib/fontmake/font_project.py
```diff
@@ -202,6 +202,8 @@ def save_otfs(
                 useProductionNames=use_production_names)
             otf.save(otf_path)
 
+            if subset is None:
+                subset = (GLYPHS_PREFIX + 'Keep Glyphs') in ufo.lib
```
jamesgk: I guess we'd also want to subset if for any glyph's lib the key `GLYPHS_PREFIX + 'Glyphs.Export'` maps to `False`.

anthrotype:
@jamesgk I just updated it so that it also runs the subsetter if there is any glyph's lib with "Glyphs.Export" set to False.

Though I don't have any such .glyphs file to test this on.

PR #113, opened 2016-07-27
font_project: move booleanOperations import inside 'remove_overalap' method

anthrotype:
If one passes `--keep-overalap` option, then `booleanOperations` should not be required by fontmake.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -22,7 +22,6 @@ import plistlib
 import re
 import tempfile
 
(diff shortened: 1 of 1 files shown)


anthrotype:
I also pushed a similar commit that moves `glyphsLib` and `mutatorMath` inside the methods which use them.

If one only wants to be able to use fontmake to build UFOs (no Glyphs files), and does not need to interpolate them, it should not be required to also install glyphsLib and mutatorMath.

PR #116, opened 2016-08-01
requirements.txt: update glyphsLib to commit 0630c67

m4rc1e:
Since glyphsLib has been updated since I submitted my first pr, I've set it to the latest.


jamesgk:
Thanks!

PR #120, opened 2016-08-03
dev_requirements.txt: install all dependencies from HEAD

anthrotype:
It could be used as a shorthand for checking if anything breaks when some dependency gets updated, before modifying the actual "requirements.txt" file (which must keep pinned-down versions).

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,0 +1,18 @@
+# for MutatorMath
+-e git+https://github.com/typesupply/fontMath.git@ufo3#egg=fontMath
+-e git+https://github.com/unified-font-object/ufoLib.git#egg=ufoLib
+
+# for ufo2ft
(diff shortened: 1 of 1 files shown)


behdad:
Nice!

Question: what does -e mean with requirements.txt?  Where is the source put?

jamesgk:
I think it's "editable", the same as specifying --user with setup.py

anthrotype:
'-e' is for "editable"
https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support

> For editable installs, the clone location by default is "<venv path>/src/SomeProject" in virtual environments, and "<cwd>/src/SomeProject" for global installs. The --src option can be used to modify this location.

anthrotype:
Same as setuptools `develop` command. `--user` is to install in the user's site-packages.

jamesgk:
Oh right, that's what I meant. Thanks

PR #121, opened 2016-08-03
requirements.txt: update all to latest HEAD

anthrotype:
I did `git rev-parse HEAD` to get the commit hash for all the dependencies, each one cloned from the respective branch's HEAD.

We desperately need a better way to handle dependencies -- and I know I keep saying this for months without actually solving anything.

I know what the solution is: tagging and versioning everything, upload everything to PyPI, so that we can specify immediate dependencies only via setup.py's "install_requires" and let pip do its magic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 install:
   # for fontmake
-  - pip install --upgrade pip setuptools Cython
+  - pip install --upgrade pip setuptools
   - pip install -r requirements.txt
   - python setup.py install
   # for fontdiff
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,18 @@
+sudo: false
 language: python
 python:
   - "2.7"
+env:
+  global:
+    - CXX=g++-4.8
+    - CC=gcc-4.8
 
-sudo: false
-os:
-  - linux
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.8
 
 install:
   # for fontmake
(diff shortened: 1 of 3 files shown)


anthrotype:
compiling the latest compreffor requires a compiler with -std=c++11 support, but the container-based Travis with Ubuntu Precise has a too old GCC. It works if we use Trusty, as that comes with 4.8.

khaledhosny:
You can get GCC 4.8 on container-based Travis, see for example https://github.com/servo/fontsan/pull/11/commits/d940413f687aa757e12e2c91f5456269b7bf7dc1.

anthrotype:
Thanks, I can try that tomorrow.

anthrotype:
I tried to do what @khaledhosny suggested, but somehow python is still using the built-in `gcc` instead of the newly installed `gcc-4.8`, even though I exported the `CC` and `CXX` variable and `distutils.sysconfig.customize_compiler` does check in `os.environ` for those variables.

I actually realised that, before we introduced the Cython wrapper for compreffor where I added the `-std=c++11` flag in setup.py, we were happily building the compreffor shared library using `make` by passing the `-std=c++0x` option, and that was working even on the Travis' Ubuntu 12.04 VM.

I opened a PR on compreffor repo; once that's merged I'm going to point fontmake's requirements.txt to the newly merged HEAD commit, and then It should work again.

anthrotype:
there was other stuff here besides the g++4.8. Shall I open a new PR for those?

anthrotype:
And in the end, I did manage to pass the CXX env variables to Travis, so the `-std=c++11` option was working now.

PR #122, opened 2016-08-03
Windows: fix "IOError: [Errno 13] Permission denied" when re-opening named temporary file

anthrotype:
on Windows we need to close the temporary file, before we can re-open it by the name... Which basically makes the whole use of NamedTemporaryFile almost pointless...

See https://bugs.python.org/issue14243

Anyway, this fixes it for now.
If we find we need to do this often, we can wrap the `try... finally` in a custom @contextmanager that does the cleanup on exit.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -28,7 +28,7 @@ from cu2qu.pens import ReverseContourPen
 from cu2qu.ufo import font_to_quadratic, fonts_to_quadratic
 from defcon import Font
 from fontTools import subset
-from fontTools.misc.py23 import tobytes
+from fontTools.misc.py23 import tobytes, UnicodeIO
 from fontTools.misc.loggingTools import configLogger, Timer
 from fontTools.misc.transform import Transform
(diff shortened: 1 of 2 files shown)


anthrotype:
rebased to fix the merge conflict.

behdad:
Do we even need to write it to a temp file?  Why not keep in memory?

anthrotype:
Maybe we don't. Feel free to close this and fix it in a different way. I'm off now.

anthrotype:
actually we do need a temp file, as glyphsLib.build_masters expects a path.

jamesgk:
Yea. We could add another function to the glyphsLib API which takes in the file contents instead; I'd be in favor of that since it would simplify things here.

anthrotype:
ok, glyphsLib can now read from either a path or an open file-like object.
I updated this PR so that fontmake passes an io.StringIO to glyphsLib instead of a path to a temporary file.

If anyone could abort the first pending Travis build, it would be nice, as it's gonna fail anyway (as I updated the requirements.txt in the follow-up commit) and it takes ages.

PR #127, opened 2016-08-05
add --subroutinize/--no-subroutinize options

anthrotype:
This PR depends on https://github.com/googlei18n/ufo2ft/pull/61

If that's merged, the fontmake default behavior would remain the same as the current one: i.e., always run compreffor if a CFF table is present.

But In addition, fontmake could gently skip running compreffor when the latter is not installed (it's still a bit tricky to compile compreffor on Windows, but I'm working on providing pre-compiled wheels for that).

On the other hand, if the '-s' ('--subroutinize') option is explicitly provided, then ufo2ft raises ImportError when compreffor is not installed.

Finally, when the '-S' ('--no-subroutinize') option is provided, fontmake tells ufo2ft to skip compreffor altogether, whether it's installed or not.

Note that on makeotf subroutinization is optional, not default.
I'm ok with leaving the current fontmake's default -- as long as it doesn't force you to install 10GB Visual Studio and monkeypatch setuptools...

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -52,7 +52,13 @@ def main():
                         action='store_true', help='Subset font using export '
                         'flags set by glyphsLib')
     group2.add_argument('--no-subset', dest='subset', action='store_false')
-    parser.set_defaults(use_production_names=None, subset=None)
+    group3 = parser.add_mutually_exclusive_group(required=False)
+    group3.add_argument('-s', '--subroutinize', action='store_true',
+                        help='Optimize CFF table using compreffor')
+    group3.add_argument('-S', '--no-subroutinize', dest='subroutinize',
+                        action='store_false')
+    parser.set_defaults(use_production_names=None, subset=None,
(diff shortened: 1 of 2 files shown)


jamesgk:
> as long as it doesn't force you to install 10GB Visual Studio and monkeypatch setuptools...

Or install a Linux VM? :smile:

anthrotype:
Yeah I know ;)

behdad:
> fontmake could gently skip running compreffor when the latter is not installed

I don't like this.  It results in hard-to-debug situations where the same fontmake command can result in vastly different fonts.  What do you think about changing default to be subroutinize?

behdad:
(imagine if we did the same for removeOverlaps...)

khaledhosny:
I was about to suggest the same (this is also true for the new ufo2ft default).

anthrotype:
Ok, let's make --subroutinize True by default. I'll make wheels for comrpeffor so one doesn't need to compile it from source.

PR #128, opened 2016-08-08
add --conversion-error option for cu2qu

anthrotype:
This controls the the `max_err_em` value that is passed on to `cu2qu.fonts_to_quadratic`.
So it should be a float number measured as a ratio of abstract font units divided by the UPEM.

The default value is the same as the current cu2qu default (which our designers think is a bit too generous): `0.0025`, or 2.5 units on a grid of 1000 UPEM.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -57,6 +57,9 @@ def main():
                         help='Optimize CFF table using compreffor')
     group3.add_argument('-S', '--no-subroutinize', dest='subroutinize',
                         action='store_false')
+    parser.add_argument('-e', '--conversion-error', type=float, default=None,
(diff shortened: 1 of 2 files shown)


jamesgk:
> which our designers think is a bit too generous

I actually agree with this, though I'm certainly no authority. I was thinking of bumping it down just a bit to 0.002. Anyways, why reproduce the default value here, when passing None to fonts_to_quadratic will use cu2qu's default?

anthrotype:
>  why reproduce the default value here
oh, that was just to print the default in the --help.
I rebased it and set it to None.

behdad:
2.5/1000 is indeed too high.  @anthrotype what number are you using?

Also, maybe we can add an option to cu2qu to do the rounding, such that the rounding error is also taken into account against the tolerance.

PR #129, opened 2016-08-09
fontmake.main: restore --subroutinize True by default

anthrotype:
See https://github.com/googlei18n/fontmake/pull/127#issuecomment-238381511


anthrotype:
This one should also be reverted: https://github.com/googlei18n/ufo2ft/pull/65

PR #146, opened 2016-09-11
requirements.txt: update booleanOperations

anthrotype:
As of version 0.5.1, booleanOperations is published on [PyPI](
https://pypi.org/project/booleanOperations) so one can `pip install booleanOperations`.


requirements.txt
```diff
@@ -11,6 +11,5 @@ git+https://github.com/googlei18n/cu2qu.git@ab3c0f82e60de746bafd2e3a9fab3a176840
 git+https://github.com/googlei18n/glyphsLib.git@5dc523f7472a9fcd4aade865a2c5f84bb272fcc6
 git+https://github.com/googlei18n/ufo2ft.git@c8826564103f0cf3c57cdd51e453e765528d12ed
 git+https://github.com/LettError/MutatorMath.git@652b075cc1cbc0b96ad6025906970e13bc77264e
---find-links https://github.com/typemytype/booleanOperations/releases
-booleanOperations==0.4
 git+https://github.com/typesupply/defcon.git@a70a27b7607a92e8b9d7181e6c58061547d19942
+booleanOperations==0.5.1
```
adrientetar: This could be >= or just booleanOperations?

jamesgk: Actually let's keep specific versions in requirements.txt for the sake of stability.

PR #148, opened 2016-09-29
Variable fonts

jamesgk:
Depends on https://github.com/googlei18n/glyphsLib/pull/75

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -33,6 +33,7 @@ from fontTools.misc.loggingTools import configLogger, Timer
 from fontTools.misc.transform import Transform
 from fontTools.pens.transformPen import TransformPen
 from fontTools.ttLib import TTFont
+from fontTools import varLib
 from ufo2ft import compileOTF, compileTTF
 from ufo2ft.makeotfParts import FeatureOTFCompiler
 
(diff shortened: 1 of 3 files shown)


Lib/fontmake/font_project.py
```diff
@@ -270,18 +263,20 @@ def run_from_glyphs(
             glyphs_source = self.preprocess(glyphs_file)
             glyphs_file = UnicodeIO(glyphs_source)
 
-        print('>> Building UFOs from Glyphs source')
-        ufos = self.build_ufos(
-            glyphs_file, interpolate, masters_as_instances, family_name)
-        self.run_from_ufos(
-            ufos, is_instance=(interpolate or masters_as_instances), **kwargs)
+        print('>> Building designspace from Glyphs source')
+        _, designspace_path, instance_data = self.build_master_ufos(
```
dougfelt: You're not requesting instances, so there's no instance_data.  This returns a shorter tuple so unpacking fails.

anthrotype:
Does this depends on the recently merged `opentype-gx` branch in fonttools?
https://github.com/fonttools/fonttools/pull/677

If so, I think it'd be better to wait until we decide what/when a fonttools release will be...

jamesgk:
Well this shouldn't affect any existing fontmake functionality, and I do think anyone working with variable fonts at this point will expect some instability.

anthrotype:
But you haven't updated the fonttools requirement in fontmake's `requirements.txt`, have you?

jamesgk:
Aha, good point.

requirements.txt
```diff
@@ -3,7 +3,7 @@ git+https://github.com/typesupply/fontMath.git@c7a898ab01045033f595f4f9b3b6422af
 git+https://github.com/unified-font-object/ufoLib.git@5a80482507e801ad4dcb46395593fb33b1e74d20
 
 # for ufo2ft
-git+https://github.com/behdad/fonttools.git@1b2af584ff34c48cfb912baeb59774517d26f411
+git+https://github.com/behdad/fonttools.git@a533a3be446f1ff759ebdb3a3f92397b1fa4de0a
```
anthrotype: to be picky, it's https://github.com/fonttools/fonttools now ;)

Lib/fontmake/font_project.py
jamesgk: The FontProject build_master_ufos method requests instances, so this should be working now.

PR #155, opened 2016-10-07
Change format of argument parsing code

jamesgk:
IMO this is more readable. Of course that's very subjective so I'm happy to close this if there are strong feelings the other way.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -26,44 +26,62 @@ def main():
         '-o', '--output', nargs='+', default=('otf', 'ttf'),
         choices=('ufo', 'otf', 'ttf', 'ttf-interpolatable', 'variable'))
 
-    parser.add_argument('-i', '--interpolate', action='store_true',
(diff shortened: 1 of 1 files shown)


dougfelt:
LGTM

moyogo:
Regarding https://github.com/googlei18n/fontmake/commit/e2e8c07b83a74fb873493f741ac14a43168553e8
```python
'''
        Args:
            glyphs_path: Path to source file.
            preprocess: If True, check source file for un-compilable content.
            family_name: If provided, uses this family name in the output.
            kwargs: Arguments passed along to run_from_designspace.
'''
```
This could be more legible as
```python
'''
        Args:
            glyphs_path: Path to source file.
            preprocess:  If True, check source file for un-compilable content.
            family_name: If provided, uses this family name in the output.
            kwargs:      Arguments passed along to run_from_designspace.
'''
```

jamesgk:
That's true. Though our style guide (which I do mostly try to follow) discourages this kind of vertical alignment since it can become a maintenance burden: http://google.github.io/styleguide/pyguide.html#Whitespace

PR #156, opened 2016-10-10
Option to interpolate layout from master binaries

jamesgk:
cc @behdad

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -18,6 +18,7 @@ from __future__ import unicode_literals
 
 import glob
 import logging
+import math
 import os
 import plistlib
 import re
@@ -180,7 +181,8 @@ class FontProject:
     def save_otfs(
             self, ufos, ttf=False, is_instance=False, interpolatable=False,
             mti_paths=None, use_afdko=False, autohint=None, subset=None,
-            use_production_names=None, subroutinize=False):
+            use_production_names=None, subroutinize=False,
+            interpolate_layout_from=None):
         """Build OpenType binaries from UFOs.
 
         Args:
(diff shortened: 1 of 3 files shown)


Lib/fontmake/font_project.py
```diff
@@ -420,6 +450,33 @@ def _output_path(self, ufo, ext, is_instance=False, interpolatable=False,
...
 
+    def _designspace_locations(self, designspace_path):
+        """Map font filenames to their locations in a designspace."""
+
+        maps = []
+        for location_list in varLib.designspace.load(designspace_path):
+            location_map = {}
+            for loc in location_list:
+                abspath = os.path.normpath(os.path.join(
+                    os.path.dirname(designspace_path), loc['filename']))
```
dougfelt: can normpath and dirname be hoisted out of this inner loop?  Or are we normalizing the loc['filename'] result as well?

jamesgk: Right, the path that gets normalized includes loc['filename'] as well.

Lib/fontmake/font_project.py
```diff
@@ -420,6 +450,33 @@ def _output_path(self, ufo, ext, is_instance=False, interpolatable=False,
...
+                location_map[abspath] = loc['location']
+            maps.append(location_map)
+        return maps
+
+    def _closest_location(self, location_map, target):
+        """Return path of font whose location is closest to target."""
+
+        dist = lambda a, b: math.sqrt(sum((a[k] - b[k]) ** 2 for k in a.keys()))
```
jamesgk: They're not normalized, but I'm not sure if they need to be. Kind of an interesting question. If all of the points would be normalized from the same space, and all we're doing is comparing the distances, can we possibly get different results normalizing vs. not?

dougfelt: I'm thinking, say axis A runs on a scale from 1-100, and B on a scale from 1-10.  A difference in A of 10 points is semantically 'equivalent' to a difference in B of 1 point, but the distance in A looks a lot larger.  This function will choose fonts with almost any B value over those with even small differences in A. So I figure these must be scaled so that differences on the various axes all contribute the 'intended' amount to the selection algorithm, even if the ranges are not exactly the same they should be close.

I guess also there are only two or maybe three axes?

dougfelt:
LGTM - whatever you think is best

Lib/fontmake/font_project.py
jamesgk: Yea but the target point would also be normalized. I'm thinking that the normalization factors cancel out in the end and so comparisons of normalized locations yield the same result.

And yep in the typical case we'd expect 2 or 3 axes. Though I think any number is possible with MutatorMath designspaces.

dougfelt: So, can you spell this out for me?  I still don't see it.

jamesgk: So I was confused about what normalization meant. But Doug and I still agreed that we don't need to normalize here as long as the ranges of the axes are the same. We can revisit this if it turns out that's not a safe assumption.

PR #158, opened 2016-10-12
Use logging module

jamesgk:
This allows the user to specify whether they want to see the build status messages or not via a --verbosity flag.

It has the side effect, however, of printing info from dependencies. In the case of MutatorMath, which otherwise runs silently for a long time, this isn't necessarily a bad thing. Though in some cases (e.g. fontTools' subsetter) the info messages can be a bit noisy.

We probably want to do something similar for glyphsLib, ufo2ft, and cu2qu.

@marekjez86: this should radically change the format of fontmake console output, but may allow us to build LGC fonts on Travis. (At least, it shouldn't time out due to MutatorMath silence.  The overall process may still take too long.)

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -87,9 +87,11 @@ def main():
                         subroutinize=True)
 
     parser.add_argument('--timing', action='store_true')
+    parser.add_argument('--verbosity', default='INFO')
     args = vars(parser.parse_args())
 
-    project = FontProject(timing=args.pop('timing'))
+    project = FontProject(timing=args.pop('timing'),
+                          verbosity=args.pop('verbosity'))
 
     glyphs_path = args.pop('glyphs_path')
     ufo_paths = args.pop('ufo_paths')
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -41,7 +41,7 @@ from ufo2ft.makeotfParts import FeatureOTFCompiler
 
 from fontmake.ttfautohint import ttfautohint
 
-timer = Timer(logging.getLogger('fontmake'), level=logging.DEBUG)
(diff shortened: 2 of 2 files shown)


behdad:
Bikeshed: most packages use --verbose with an option argument (like --verbose=4)

jamesgk:
You mean, rename the flag from verbosity to verbose? It does take an argument, but it's the name (INFO, DEBUG, WARNING etc.) rather than a number.

jamesgk:
(Sorry, "flag" isn't the right word here, is it.)

anthrotype:
Using logging is a great idea, I've been wanting to do that for a while. Good you picked it up first!
I haven't looked at the code yet, I'll do that soon. About the --verbose flag, what if you could specify it multiple times to signal a greater degree of verbosity? E.g. similar to what I did for compreffor

behdad:
> You mean, rename the flag from verbosity to verbose?

Yes, that was my suggestion.

> I haven't looked at the code yet, I'll do that soon. About the --verbose flag, what if you could specify it multiple times to signal a greater degree of verbosity?

That's bad bad bad.  Please don't do that!

behdad:
In general I suggest keeping things close to GNU style:
https://www.gnu.org/prep/standards/

anthrotype:
Well, many command line tools I've come across do that, but  actually I don't really mind either way as long as it's clear and consistent :)

anthrotype:
I actually don't see why the "additive" --verbose, as in `-vvv`, would be so "bad bad bad"...
Even `pip` command, arguably one of the most "pythonic" of all,  does exactly that, among several others:

```
$ pip --help
[...]
-v, --verbose               Give more output. Option is additive, and can be used up to 3 times.
```

btw, I'd also prefer "verbose" over "verbosity" -- but I can see why you opted for the latter, since you require an argument with the logging level name. --verbose="CRITICAL"  does not really make much sense.

PR #160, opened 2016-10-13
Fix for verbose KeyError

traviskochel:
On a clean install, when I run anything, it throws a `KeyError: 'verbosity'`. For example `fontmake -o ttf-interpolatable -m myfont.designspace`, I get this:

screen shot 2016-10-13 at 1 26 14 pm

It seems like there was a leftover `verbosity` reference that needed to be renamed, from this commit: https://github.com/googlei18n/fontmake/commit/bfc7ef692937020a13f3995c1462f6c2b1fc60df

I'm not sure if this solution is the best fix, or if there are more references, but it works for me now.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -91,7 +91,7 @@ def main():
     args = vars(parser.parse_args())
 
     project = FontProject(timing=args.pop('timing'),
-                          verbosity=args.pop('verbosity'))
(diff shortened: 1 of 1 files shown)


traviskochel:
I signed it!

jamesgk:
Oops, didn't see this. Should have been fixed by d53cf28b457ffb5e6330261ed3d893e2f357f2ca.

PR #162, opened 2016-10-17
font_project.py: log the font/glyph names when booleanOperations fail

anthrotype:
Sometimes booleanOperations may fail because the input contours are invalid for clipping (e.g. see https://github.com/googlei18n/fontmake/issues/161)

With https://github.com/typemytype/booleanOperations/pull/37, we can catch a `BooleanOperationsError`, and let users know where to find the invalid path, so they can fix it.

Also, we can use `booleanOperations.union` function directly, instead of instantiating a `BooleanOperationManager` instance, as the `union` method is static anyway.

This PR depends on the above PR being merged first.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -95,15 +95,20 @@ class FontProject:
     @timer()
     def remove_overlaps(self, ufos):
         """Remove overlaps in UFOs' glyphs' contours."""
-        from booleanOperations import BooleanOperationManager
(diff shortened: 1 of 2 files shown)


jamesgk:
Needs an update to booleanOperations requirement?

anthrotype:
yep. I would like to tag a 0.6.0 release tomorrow.
But the PR was merged so maybe in the meantime we could point to HEAD.

PR #163, opened 2016-10-18
update requirements.txt to download packages from PyPI

anthrotype:
download the wheel from PyPI, so installation is quicker.


anthrotype:
ping @jamesgk

PR #176, opened 2016-11-01
Add end-to-end test from designspace source

jamesgk:


behdad:
Excellent.  Thanks James!

PR #179, opened 2016-11-03
More robust building of variable font

jamesgk:
Before, fontmake would assume master UFOs were found in a master_ufo
directory. Now hopefully they can be just about anywhere.


dougfelt:
LGTM

PR #200, opened 2016-12-01
Add --use-kern-writer and --use-mark-writer options

twardoch:
Add `--use-kern-writer` and `--use-mark-writer` options that allow me to override the `ufo2ft` built-in `KernFeatureWriter` and `MarkFeatureWriter` classes with my custom ones. 

I made one such class in [`feaLab`](https://github.com/twardoch/feaLab) which basically does nothing. This allows me to use `fontmake` to build [`VotoSerifGX-OFL`](https://github.com/twardoch/varfonts-ofl/tree/master/VotoSerifGX-OFL) where `fontmake` currently fails because it cannot merge OTL tables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
 dist
+src
 *.pyc
 *.egg-info/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 build
 dist
-src
 *.pyc
 *.egg-info/
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -19,82 +19,90 @@ from fontmake.font_project import FontProject
 
 def main():
     parser = ArgumentParser()
-    parser.add_argument('-g', '--glyphs-path')
-    parser.add_argument('-u', '--ufo-paths', nargs='+')
-    parser.add_argument('-m', '--mm-designspace')
-    parser.add_argument(
+    inputGroup = parser.add_argument_group(title='Input arguments')
+    inputGroup.add_argument('-g', '--glyphs-path',
+        help='Path to .glyphs source file')
+    inputGroup.add_argument('-u', '--ufo-paths', nargs='+',
+        help='Paths to UFO files, should be in "master_ufo/" subfolder')
+    inputGroup.add_argument('-m', '--mm-designspace',
+        help='Path to .designspace file while the UFOs should be in "master_ufo/" subfolder')
+
+    outputGroup = parser.add_argument_group(title='Output arguments')
+    outputGroup.add_argument(
         '-o', '--output', nargs='+', default=('otf', 'ttf'),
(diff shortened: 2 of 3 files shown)


twardoch:
Rationale: Generating `kern`, `mark` and `mkmk` features is quite a specific operation. There may be many strategies to do that, and lots of potential optimizations, so I expect some external libraries to emerge which do it, and one could even have custom generators for a specific project (for example that would insert additional contextual kerning lookups). `ufo2ft` already provides a mechanism to override its built-in generators, and this patch exposes this mechanism to `fontmake`.

.gitignore
```diff
@@ -1,4 +1,5 @@
 build
 dist
+src
```
anthrotype: why this?

twardoch:
I’ve also beautified the usage info of `fontmake` a bit.

.gitignore
twardoch: Because of `dev_requirements.txt`. When I do `pip install -r dev_requirements.txt`, Python will create a `src/` subfolder in the repo and then clone all the requirements and install them in developer mode from there.

anthrotype:
Hey Adam! Thanks, this is quite useful 👍

anthrotype:
I wonder whether the changes to the argument parsing should go to a separate pull request, maybe?
Just to make code review easier.

moyogo:
Nice. I started working on something related to this in ufo2ft.
I wanted to be able to give any feature writer (in the order they should write lookups) instead of only kern and mark/mkmk feature writers.

.gitignore
anthrotype: When pip checks out editable projects and you are running in a virtualenv, pip places them by default in `<venv path>/src`. For global installs, the default is `<current dir>/src`.

You can change pip behaviours by passing the `--src` option:
`pip install --src <dir> -r dev_requirements.txt`

And you can configure your local copy of the repo to ignore some files by doing, e.g.:

```
echo "src/" >> .git/info/exclude
```

twardoch:
Do I have to create new branches for each potential pull request?

anthrotype:
Yes, each pull request has its own branch. But you could keep this one as a single PR, I don't mind.

twardoch:
I'll figure out separate branches for future PRs but go ahead and merge now as is. I've reverted the .gitignore hack.

twardoch:
@moyogo For arbitrary steps, I think we should look into some lightweight control files (maybe not Makefile unless we have decent generators). Because after all in a process like this, one wants hook to exist at different steps of the process: when writing from .glyphs or other formats.to ufo, when writing from ufos to ttfs, and when merging the ttfs. Currently fontmake is a bit of a hacky tool. I'm not am expert on build processes myself but it'd be great to find some method that combines relative simplicity with flexibility, and ideally uses standard software, and is cross platform.

behdad:
> it'd be great to find some method that combines relative simplicity with flexibility, and ideally uses standard software, and is cross platform.

That has been my longterm goal for fontmake.  We'll get there some day :D.

PR #207, opened 2016-12-07
follow up on PR #200

anthrotype:
hey Adam @twardoch 
I reviewed your PR #200 and added a few commits on top of it.

- I renamed the options to `--kern-writer-module` and `--mark-writer-module`; I think it's better as they take a module name argument, while `--use-...` seems to imply a boolean value.

- the argument parser passes the class objects themselves rather than the module names; this also gives the chance to raise a parser.error when the module can't be imported or does not contain a class with the expected name. I think it's more reasonable to raise here, instead of just log a warning.

- I kept the grouping of the arguments, and only made cosmetic changes here and there.

Please take a look, thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build
 dist
+src
 *.pyc
 *.egg-info/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 build
 dist
-src
 *.pyc
 *.egg-info/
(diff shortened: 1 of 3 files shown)


anthrotype:
@twardoch I guess to appease googlebot you'll have to confirm you're okay with this. Sorry...

twardoch:
confirm

PR #208, opened 2016-12-07
Decompose in TTF build

jamesgk:
Biggest change is that we decompose any glyph with contours when compiling TTFs. Also contains minor logging and doc additions.


dougfelt:
So does this mean overlap removal will happen so the ttfs will be clean in this respect?

jamesgk:
The glyphs with only components might still have overlaps, but there shouldn't be any overlaps of contours.

PR #213, opened 2016-12-09
Update README.md

davelab6:
Add more notes about installing, using and updating

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,21 +6,32 @@ This library provides a wrapper for several other Python libraries which
 together compile fonts from various sources (.glyphs, .ufo) into binaries (.otf,
 .ttf).
 
-### Install
+### Initial Installation
 
-Create new virtual environment (optional, but *recommended*):
+First, create new virtual environment. 
+This is optional, but **recommended**:
+
 ```bash
 python -m pip install --user virtualenv  # install virtualenv if not available
 python -m virtualenv env  # create environment named 'env' in current folder
 source env/bin/activate  # activate the environment (run `deactivate` to exit)
 ```
 
+Clone this repo, and enter it:
+
+```bash
+git clone git@github.com:googlei18n/fontmake.git;
+cd fontmake;
+```
+
 Install fontmake's dependencies:
+
 ```bash
 pip install -r requirements.txt
 ```
 
-Install fontmake:
+Install fontmake itself:
+
 ```bash
 pip install -e .  # `-e` is for "editable" mode, only required for developers
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,3 +50,32 @@ Use `-h` to see options for specifying different types of input and output.
...
+
+```bash
+source env/bin/activate  # activate the environment (run `deactivate` to exit)
+```
+
+Enter this repo, and pull this repo:
+
+```bash
+cd fontmake;
+git clone git@github.com:googlei18n/fontmake.git;
```
anthrotype: perhaps here you meant
```
git pull
```

README.md
```diff
@@ -38,3 +50,32 @@ Use `-h` to see options for specifying different types of input and output.
...
+```
+
+Reinstall fontmake's dependencies to their latest versions:
+
+```bash
+pip install -r requirements.txt
```
anthrotype: Since the section is about "updating", you may wish to do `pip install --upgrade -r requirements.txt` (or `-U`).

If you don't specify `--upgrade`, pip will say `Requirements already satisfied` for those dependencies which are currently installed from their git repositories but which do not regularly modify their package version.

E.g. MutatorMath tells pip that its package version is 1.8, but that hasn't changed in years despite many changes to the code. So pip does not update because it must assume the requirement is already satisfied (i.e. you already have a MutatorMath==1.8 installed)

For fonttools, since the version string is computed from git metadata, pip is still able to detect that the package version is different, if the commit hash in the requirements.txt changes.

README.md
```diff
@@ -38,3 +50,32 @@ Use `-h` to see options for specifying different types of input and output.
...
+```
+
+Reinstall fontmake itself:
+
+```bash
+pip install -e .  # `-e` is for "editable" mode, only required for developers
```
anthrotype: If you installed "editable" you don't need to re-run `pip install -e .`; `git pull` does the job.
If you installed without editable, then yes, you need to re-do `pip install .`

anthrotype:
I don't know if it's necessary to have two sub-sections for "initial installation" and "updating"...

Dave, if you wait a couple of days (once glyphsLib, MutatorMath and fontmake itself are uploaded to PyPI), your installation instructions will simply be:

> To install/update fontmake along with its dependencies, do pip install -U fontmake, preferably from inside a virtual environment.

davelab6:
> I don't know if it's necessary to have two sub-sections for "initial installation" and "updating"...

The googlei18n team specifically requested it in an internal fontmake workshop yesterday :)

davelab6:
> pip install -U fontmake

That will be awesome!

davelab6:
I updated the README with all the info in your comments - please wait a couple of days until the new package is released :)

anthrotype:
Dave, I had to change the README to reStructuredText, so it can be shown on PyPI as well.
Do you mind applying the same changes but on README.rst?
Thanks

davelab6:
No worries - I'm vaguely recalling that MD was supported by PyPi, but rST
is fine :)

anthrotype:
No, I'm afraid MD is not supported yet on current PyPI:
https://github.com/pypa/pypi-legacy/issues/148

PR #227, opened 2017-01-20
Add designspace test with axis element

jamesgk:
MutatorMath requirement needs to be updated to latest, and then hopefully this will pass.


LettError:
Is this what you need: https://github.com/LettError/MutatorMath/releases/tag/2.01

anthrotype:
@LettError yes, thank you Erik!

@jamesgk if you rebase on master, the test should now pass.

Issue #199, opened 2016-11-30
add a instance folder for GX fonts


xiangyexiao:
VF fonts are currently created in master_ufo, would be more convenient to put them in a separate folder.

marekjez86:
this is already done

marekjez86:
variable_ttf folder

PR #266, opened 2017-03-03
Move variable font output to separate folder, add test cases to fontmake

zjusbo:
#199

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -29,7 +29,7 @@ from cu2qu.pens import ReverseContourPen
 from cu2qu.ufo import font_to_quadratic, fonts_to_quadratic
 from defcon import Font
 from fontTools import subset
-from fontTools.misc.py23 import tobytes, UnicodeIO
+from fontTools.misc.py23 import tobytes, UnicodeIO, basestring
 from fontTools.misc.loggingTools import configLogger, Timer
 from fontTools.misc.transform import Transform
 from fontTools.pens.transformPen import TransformPen
@@ -116,7 +116,7 @@ class FontProject(object):
             # if the transformation has a negative determinant, it will reverse
             # the contour direction of the component
             xx, xy, yx, yy = transformation[:4]
-            if xx * yy - xy * yx < 0:
+            if xx*yy - xy*yx < 0:
                 pen = ReverseContourPen(pen)
 
             component.draw(pen)
@@ -463,13 +463,7 @@ class FontProject(object):
             Returns ufo directly if ufo is a string. This case happens when
             generating variable font, where ufo stores the output font name.
         """
-        l = [str]
-        try:
-            # Python 3+ does not have unicode keyword.
-            l.append(unicode)
-        except NameError:
-            pass
-        if type(ufo) in l :
+        if isinstance(ufo, basestring) :
             return ufo
         return '%s-%s' % (ufo.info.familyName.replace(' ', ''),
                           ufo.info.styleName.replace(' ', ''))
@@ -487,7 +481,7 @@ class FontProject(object):
             Return:
                 output directory string.
         """
-
+        # FIXME? Use user configurable destination folders.
         dir_prefix = 'instance_' if is_instance else 'master_'
         dir_suffix = '_interpolatable' if interpolatable else ''
         dir_suffix = dir_suffix + '_variable' if is_variable else dir_suffix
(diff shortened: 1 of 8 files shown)


zjusbo:
Some questions:

1. How should we test the python version in our code base. I used try/except. The other option is check sys.version_info_

2. What does 'interpolatable' mean, can this parameter coexist with 'variable'?

   def _output_dir(self, ext, is_instance=False, interpolatable=False,
                    autohinted=False, is_variable=False):

Lib/fontmake/font_project.py
```diff
@@ -453,30 +455,52 @@ def run_from_ufos(
...
+            Returns ufo directly if ufo is a string. This case happens when
+            generating variable font, where ufo stores the output font name.
+        """
+        l = [str]
+        try:
+            # Python 3+ does not have unicode keyword.
+            l.append(unicode)
+        except NameError:
+            pass
+        if type(ufo) in l :
```
anthrotype: here you can do this: 

```python
from fontTools.misc.py23 import basestring

if isinstance(ufo, basestring):
```

anthrotype:
> How should we test the python version in our code base

in general, checking `sys.version_info` is the right way. You can also import `PY2` and `PY3` constants from `fontTools.misc.py23` module, which come from sys.version_info anyway.

> interpolatable

This means that only master TTFs are built, but not the final variable font. The `ttf-interpolatable` format option was introduced before the `variable` one, and at that time the only way to make a variable font was to first make the master interpolatable TTFs, and then run fontTools.varLib.main script on them.

---

Bo, I actually don't think we need an extra folder for storing the output variable font.
It's never going to be more than one file anyway, what's the point in creating a folder then?

If you also find confusing that the generated *-VF.ttf ends up next in `master_ufo` folder, then we can just change that and save the variable font in the current folder.

BTW, I've always found fontmake's use of hard-coded destination folders ("master_ufo", "instance_ttf") not ideal. I wish these could be customized somehow. The `-o` option for specifying the output _format_ instead of the _destination_ is quite unusual, I believe. But these are probably different issues.

verbosus:
I have a patch for specifying the output folder in general (if I rememer right I added `-d --directory` since `-o --output` was already used), give me some time to clean it up.

setup.py
```diff
@@ -43,6 +43,7 @@
         "MutatorMath>=2.0.2",
         "defcon>=0.2.1",
         "booleanOperations>=0.6.4",
+        "mock>=2.0.0",
```
anthrotype: I like that you used the `mock` module to write the unit tests.

However, if anything,`mock` is only a _test_ requirement, not an _install_ requirement, so it should not be in `install_requires`.

Similarly, the `requirements.txt` file should not contain `mock` either, as that's used for specifying install requirements with pinned-down version; if we want the tests to be run with fixed versions of the test requirements, we could create a separate `test-requirements.txt` file or similar.

test/test_arguments.py
```diff
@@ -0,0 +1,83 @@
...
+from mock import patch
+import mock
+from fontTools.ttLib import TTFont
+from fontmake.font_project import FontProject
+import fontmake.__main__ as entry
+import fontTools
+
+
+class TestFunctionsAreCalledByArguments(unittest.TestCase):
+  @patch('fontmake.font_project.FontProject.run_from_glyphs')
```
anthrotype: we use four spaces indentation everywhere else; this file uses two.

test/test_output.py
```diff
@@ -18,7 +18,6 @@
 
 from fontTools.ttLib import TTFont
 
-
```
anthrotype: removing this line is not necessary. Keeping two empty lines between module-scope classes or function definitions is recommended in PEP8.

Lib/fontmake/font_project.py
zjusbo: Thanks Cosimo. I'll utilize misc.py23 more.

setup.py
zjusbo: I created a test-requriements.txt file and probably I'm changing our test engine config.

test/test_arguments.py
zjusbo: Thans Cosimo. Done.

test/test_output.py
zjusbo: Oh Thanks for pointing the PEP8 style guide. It's worth reading and I looked though it. I made some style changes in the codebase based on its recommendation.

zjusbo:
@anthrotype Is it posible for user to create multiple VF fonts by running a script? I was thinking in @marekjez86 's use case. He may create many VF fonts at one time. So probably dumping them into a single folder seems reasonable? Please correct me if I'm wrong.

@verbosus Thanks for doing it! adding '-d' in command line argument would be more user friendly.

PR #269, opened 2017-03-07
make interpolate/variable mutually exclusive; save VF to "variable_ttf"

anthrotype:
This is a follow-up on PR #266.

The `-i/--interpolate` option is used to make instance UFOs with MutatorMath and shouldn't be used along with `-o variable` output; `fontTools.varLib` expects interpolatable _master_ TTFs (plus designspace), not _instances_.

Therefore, for the `build_variable_font` function, the argument `is_instance` or `interpolatable` don't make any sense and I removed them.

I also changed the default folder where variable fonts gets generated to "variable_ttf".
In the previous PR #266, @zjusbo used "master_ttf_variable", but I believe neither the 'master' nor 'instance' attributes qualify really well a variable font, which is... both!

Finally, since when generating a variable font we require a designspace file (or a glyphs file from which we generate a designspace file), it makes sense to stop earlier (at the argument parsing step), instead of failing much later on.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -161,12 +161,21 @@ def main(args=None):
              '%(choices)s. Default: INFO')
     args = vars(parser.parse_args(args))
 
-    project = FontProject(timing=args.pop('timing'),
-                          verbose=args.pop('verbose'))
-
     glyphs_path = args.pop('glyphs_path')
     ufo_paths = args.pop('ufo_paths')
     designspace_path = args.pop('mm_designspace')
+
+    if 'variable' in args['output']:
+        if not (glyphs_path or designspace_path):
+            parser.error(
+                'Glyphs or designspace source required for variable font')
+        if args['interpolate']:
+            parser.error(
+                '"interpolate" argument invalid for variable font')
+
+    project = FontProject(timing=args.pop('timing'),
+                          verbose=args.pop('verbose'))
+
     if glyphs_path:
         project.run_from_glyphs(glyphs_path, **args)
         return
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -169,9 +169,11 @@ def main(args=None):
         if not (glyphs_path or designspace_path):
             parser.error(
                 'Glyphs or designspace source required for variable font')
-        if args['interpolate']:
-            parser.error(
-                '"interpolate" argument invalid for variable font')
+        for argname in ('interpolate', 'masters_as_instances',
+                        'interpolate_binary_layout'):
+            if args[argname]:
+                parser.error('--%s option invalid for variable font'
+                             % argname.replace("_", "-"))
 
     project = FontProject(timing=args.pop('timing'),
                           verbose=args.pop('verbose'))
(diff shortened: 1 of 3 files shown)


Lib/fontmake/font_project.py
```diff
@@ -20,16 +20,18 @@
...
 import os
-import plistlib
-import re
 import tempfile
-from io import open
+try:
+    from plistlib import load as readPlist  # PY3
+except ImportError:
+    from plistlib import readPlist  # PY2
+
```
zjusbo: nit: How about moving this to misc.py23?

Lib/fontmake/__main__.py
```diff
@@ -161,12 +161,23 @@ def main(args=None):
...
-                          verbose=args.pop('verbose'))
-
     glyphs_path = args.pop('glyphs_path')
     ufo_paths = args.pop('ufo_paths')
     designspace_path = args.pop('mm_designspace')
+
+    if 'variable' in args['output']:
+        if not (glyphs_path or designspace_path):
+            parser.error(
+                'Glyphs or designspace source required for variable font')
```
zjusbo: I'm not sure but why can't we generate VF from ufo file?
IIUC, glyphs are converted to ufo file, then ufo file is compiled to binary.

Lib/fontmake/font_project.py
zjusbo: Oh, if it's just used in fontmake, I'm ok it's being there. We'd better create a misc.py23 package for fontmake afterwards.

zjusbo:
Thanks for fixing it. LGTM.

Lib/fontmake/font_project.py
moyogo: Or this could be `from ufoLib.plistlib import readPlist`.

Lib/fontmake/__main__.py
moyogo: A UFO3 only contains one font. You’d need at least two fonts and a designspace, or a Glyphs file to make a VF.

zjusbo: Thanks Denis. I understand we need at least two fonts, or say masters, to determine axis plane on which the fonts can vary. But what is a designspace? I guess it is the span in which the the VF can be transformed? 

Say, if we have two masters, the designspace is 1D, a segment line.
if we have three masters (not on a same axis), the designspace is a 2D plane (probably a triangle, rectangle, eclipse, etc.).
And [this](https://cdn-images-1.medium.com/max/1600/1*jNGTUj2PX2PXHv6sYUrXCg.png) is a designspace for 3D, a cube.

Is it because we need to avoid intolerant distortions or unreasonable glyphs shapes so that we use designspace to restrict the variation of a font?

moyogo: Yeah, the info in the UFOs is not sufficient. Even if they have openTypeOS2WeightClass and openTypeOS2WidthClass, that’s not enough to make a VF.

The designspace is a file (see https://github.com/LettError/designSpaceDocument or https://github.com/LettError/MutatorMath or https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/varLib/designspace.py) which fontTools.varLib uses to write the OpenType design-variation space information and which MutatorMath uses to interpolate instance UFOs.

PR #277, opened 2017-03-17
[travis] update pip to install from precompiled manylinux1 wheels

anthrotype:
The pip version installed by default on the Travis python is too old and does not support installing from binary 'manylinux1' wheels.
Upgrading pip should speed up the CI builds, as we can install pyclipper and compreffor from the pre-compiled wheel packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.5"
 
 install:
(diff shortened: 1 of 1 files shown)


anthrotype:
build time reduced by more than one minute

PR #290, opened 2017-04-12
Copy MTI feature files into master UFOs

brawer:
After this change, the master UFOs are self-contained in the sense
that no external files are needed to compile them, even in workflows
that use Monotype (MTI) feature files instead of Adobe feature files.

We still need to change ufo2ft so it actually recognizes these newly
embedded files, but this can be done independently.

https://github.com/googlei18n/fontmake/issues/289

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -26,7 +26,10 @@ try:
     from plistlib import load as readPlist  # PY3
 except ImportError:
     from plistlib import readPlist  # PY2
-import xml.etree.ElementTree as etree
+try:
+    import xml.etree.cElementTree as etree
+except ImportError:
+    import xml.etree.ElementTree as etree
 
 
 from cu2qu.pens import ReverseContourPen
@@ -90,6 +93,8 @@ class FontProject(object):
                 os.mkdir(data_dir)
             target_dir = os.path.join(
                 data_dir, 'com.github.googlei18n.ufo2ft.mtiFeatures')
+            if os.path.exists(target_dir):
+                shutil.rmtree(target_dir)
             os.mkdir(target_dir)
             key = master.attrib['filename'].rstrip('.ufo')
             for table, path in mti_paths[key].items():
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -20,11 +20,13 @@ import glob
 import logging
 import math
 import os
+import shutil
 import tempfile
 try:
     from plistlib import load as readPlist  # PY3
 except ImportError:
     from plistlib import readPlist  # PY2
+import xml.etree.ElementTree as etree
 
 
 from cu2qu.pens import ReverseContourPen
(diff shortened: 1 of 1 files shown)


Lib/fontmake/font_project.py
```diff
@@ -20,11 +20,13 @@
 import logging
 import math
 import os
+import shutil
 import tempfile
 try:
     from plistlib import load as readPlist  # PY3
 except ImportError:
     from plistlib import readPlist  # PY2
+import xml.etree.ElementTree as etree
```
anthrotype: For compatiblity with Python 2, I'd suggest here you try to import the faster C implementation `cElementTree` first, and if that fails (on platforms that might not have the `_elementtree` C extension module) use the pure Python one. Python 3 does that automatically when you `import xml.etree.ElementTree`.

```python
try:
	import xml.etree.cElementTree as etree
except ImportError:
	import xml.etree.ElementTree as etree
```

Lib/fontmake/font_project.py
```diff
@@ -62,15 +64,44 @@ def __init__(self, timing=False, verbose='INFO'):
...
+        master_dir = os.path.dirname(designspace_path)
+        designspace = etree.parse(designspace_path)
+        for master in designspace.findall('sources/source'):
+            master_path = os.path.join(master_dir, master.attrib['filename'])
+            data_dir = os.path.join(master_path, 'data')
+            if not os.path.exists(data_dir):
+                os.mkdir(data_dir)
+            target_dir = os.path.join(
+                data_dir, 'com.github.googlei18n.ufo2ft.mtiFeatures')
+            os.mkdir(target_dir)
```
anthrotype: would this `os.mkdir(target_dir)` raise `FileExistsError` (or `OSError` 17) if the `target_dir` already exists?

Lib/fontmake/font_project.py
```diff
@@ -62,15 +64,44 @@ def __init__(self, timing=False, verbose='INFO'):
...
+            for table, path in mti_paths[key].items():
+                src_path = os.path.join(mti_dir, path)
+                target_path = os.path.join(target_dir, '%s.mti' % table)
+                shutil.copyfile(src_path, target_path)
+        # If we have MTI sources, any Adobe feature files derived from
+        # the Glyphs file should be ignored. If features.fea was generated,
+        # we delete it here because it only contains junk information anyway.
+        adobe_features_path = os.path.join(master_path, 'features.fea')
+        if os.path.exists(adobe_features_path):
+            os.remove(adobe_features_path)
```
anthrotype: it's a pity that glyphsLib has to write the `features.fea` (it contains the GDEF, for example) only for it to be deleted here.. But it's fine.

Lib/fontmake/font_project.py
brawer: Done.

Lib/fontmake/font_project.py
brawer: Done.

Lib/fontmake/font_project.py
brawer: Agree.

PR #292, opened 2017-04-14
Clean up workflow for Monotype feature files

brawer:
Instead of injecting the Monotype feature files late into the
pipeline, we now embed them into UFO files so that the UFOs are fully
self-contained. This change also cleans up the way how the data
gets embedded into the UFO; instead of changing files underneath
defcon, fontmake now calls the defcon data APIs.

The handling of Monotype feature files was added to ufo2ft in
version 0.4.3, so fontmake now requires that version.

https://github.com/googlei18n/fontmake/issues/289

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -72,41 +72,33 @@ class FontProject(object):
         import glyphsLib
         master_dir = self._output_dir('ufo')
         instance_dir = self._output_dir('ufo', is_instance=True)
-        _, designspace_path, instances = glyphsLib.build_masters(
+        masters, designspace_path, instances = glyphsLib.build_masters(
             glyphs_path, master_dir, designspace_instance_dir=instance_dir,
             family_name=family_name)
         if mti_source:
-            self.add_mti_features_to_master_ufos(mti_source, designspace_path)
+            self.add_mti_features_to_master_ufos(
+                mti_source, designspace_path, masters)
         return designspace_path, instances
 
     @timer()
-    def add_mti_features_to_master_ufos(self, mti_source, designspace_path):
+    def add_mti_features_to_master_ufos(self, mti_source, designspace_path,
+                                        masters):
+        mti_dir = os.path.dirname(mti_source)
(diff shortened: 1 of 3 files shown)


brawer:
@anthrotype, how to tell pip about ufo2ft 0.4.3? Currently, pip install fails because it doesn’t know about ufo2ft 0.4.3 yet.

requirements.txt
```diff
@@ -1,7 +1,7 @@
 fonttools==3.9.2
 cu2qu==1.1.1
 glyphsLib==1.6.0
-ufo2ft==0.4.2
+ufo2ft==0.4.3
```
anthrotype: I never pushed a 0.4.3 tag so there's no package with that version on pypi. Sorry for the misunderstanding.
The next ufo2ft version will be 0.5.0. It will be live only after we cut a release and push the tag.

Note that in this case, it is not sufficient to bump the ufo2ft version in "requirements.txt". You also need to manually bump the *minimum required version* of ufo2ft in the setup.py `install_requires`.
Otherwise, if one attempts to `pip install fontmake` in an environment which already contains a version of "ufo2ft>=0.4.2", the latter will not be updated automatically, since pip will think that the requirement is already satisfied.

brawer: Done; cut a release of ufo2ft (0.5.0) and using that now.

Lib/fontmake/font_project.py
```diff
@@ -298,7 +290,7 @@ def save_otfs(
                     GLYPHS_PREFIX + "Don't use Production Names")
             compiler_options = dict(
                 featureCompilerClass=fea_compiler,
-                mtiFeaFiles=mti_paths[name] if mti_paths is not None else None,
+                mtiFeaFiles=None,
```
brawer: Sure; done.

anthrotype:
Before the next fontmake release you may also want to set minimum required mutatormath to 2.0.4, Erik fixed some issue there.

anthrotype:
sorry Sascha, I bumped mutatorMath myself on master, then I used github's merge conflict resolution interface to fix up. It's actually quite cool.

PR #294, opened 2017-04-20
Update glyphsLib from 1.6.0 to 1.7.0

brawer:
Notable changes:

* designspace documents now contain `<axes>` for variation fonts

* the value range of the width axis now conforms to the OpenType spec

* font instances can now specify custom PostScript font names
  by setting the `postscriptFontName` attribute in Glyhps

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 fonttools==3.10.0
 cu2qu==1.1.1
-glyphsLib==1.6.0
+glyphsLib==1.7.0
 ufo2ft==0.5.0
 MutatorMath==2.0.4
 defcon==0.3.2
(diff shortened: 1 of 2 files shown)


brawer:
@anthrotype, how to tell pip about the new glyphsLib release?

anthrotype:
Once you tag a new release and the CI bots upload the distribution files to the Python Package Index, the pip will be able to download and install them.
GlyphsLib 1.7.0 seems to be live now at https://pypi.org/project/glyphsLib/
Try to restart the failing Travis build and see if that works now.

brawer:
Indeed, apparently pip just needed some time to catch up. However, when upgrading to glyphsLib 1.7.0, the [fontmake tests are failing because of a crash in varLib](https://travis-ci.org/googlei18n/fontmake/jobs/223860046), so this needs more debugging.

anthrotype:
https://github.com/googlei18n/glyphsLib/issues/170

PR #313, opened 2017-05-04
fix FDKFeatureCompiler and --mti-source

anthrotype:
- Fix AttributeError with new ufo2ft  FeatureCompiler (v0.5.0)
- make --use-afdko and --mti-source options mutually exclusive
- remove unused arguments and update some related docstrings

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -183,7 +183,7 @@ def main(args=None):
         project.run_from_glyphs(glyphs_path, **args)
         return
 
-    exclude_args(parser, args, ['family_name'], 'Glyphs')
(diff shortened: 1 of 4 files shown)


brawer:
Thanks, looks good!

anthrotype:
yeah but I need to fix the failing tests :-P

anthrotype:
It seems like I need to use the new "Variation Font Origin" custom parameter to make the InterpolationLayoutTest pass. The latter only has two masters, Light and Bold, and no intermediate "Regular" master at 400, so the assertion error "no base master found" is triggered by varLib, as glyphsLib (v1.7.1) still uses 400 as hardcoded default.

I just tagged glyphsLib 1.7.2, and will update the fontmake requirements.txt accordingly https://github.com/googlei18n/glyphsLib/releases/tag/v1.7.2

PR #356, opened 2017-09-23
Pass MTI sources to run_from_designspace 

punchcutter:
Until now we have only supplied MTI OT sources for masters.  Binary interpolation is then done on the resulting fonts to create the instances. If MTI sources exist for all instances then we should apply them instead of relying on interpolation. 

Old way:
1. `fontmake -g NotoSerifKannada-MM.glyphs --mti-source NotoSerifKannada-MM.plist --no-production-names`
2. `fontmake -g NotoSerifKannada-MM.glyphs -i --interpolate-binary-layout --no-production-names`

New way: 
1. `fontmake -i -g NotoSerifKannada_MM.glyphs --mti-source NotoSerifKannada-MM.plist --no-production-names `

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-fonttools==3.15.1
-cu2qu==1.2.0
+fonttools==3.19.0
+cu2qu==1.3.0
 glyphsLib==1.8.0
 ufo2ft==0.6.2
-MutatorMath==2.0.4
-defcon==0.3.4
+MutatorMath==2.1.0
+defcon==0.3.5
 booleanOperations==0.7.1


punchcutter:
Actually, I'm not so sure we need this, but it could be possible. Binary interpolation is working fine so maybe we don't ever need this.

anthrotype:
@punchcutter I actually the way you proposed better than the current one, as it avoids having to call fontmake twice when one wants to build instances with MTI sources.

punchcutter:
@anthrotype Yes, that's what I was thinking originally, but then I thought about it more and for the current files that use MTI sources there are only master sources. In some cases there are 6 or 8 masters and 36 instances. So we'd have to create MTI sources for all of those instances to do it this way and that's a ton of files to create. I think having the option to do it this way isn't bad, but binary interpolation for the instances is working pretty well.

anthrotype:
Yes, the binary GPOS interpolation is ok. What I mean is that we should try to avoid calling fontmake twice, once to build the master fonts with `--mti-source` option, and then again for the instances with `-i --interpolate-binary-layout` options. Not only this looks a bit confusing to me, but we are also creating the master UFOs twice, the first time to build the master TTFs, the second time to generate the (interpolated?) instance UFOs from which the binary instances are built.

I'm thinking that we should perhaps change it so that any of these scenarios are supported:

1) MTI sources exist only for the masters and one wishes to only build "master" OTFs or TTFs; instances correspond one to one to the masters, and are there only to make Glyphs.app happy (e.g. the Noto Urdu Nastaliq has two masters which are deliberately not interpolatable; see https://github.com/googlei18n/noto-source/pull/89):
```
fontmake -g Font.glyphs --mti-source Font.plist
```

2) MTI sources exist for all the instances and no binary GPOS interpolation is thus required
```
fontmake -g Font.glyphs --interpolate --mti-source Font.plist
```

3) if MTI sources exist only for the masters (or exists for some but not all the instances?) and binary layout interpolation is needed for the instances (that don't already have layout tables)
```
fontmake -g Font.glyphs -i --mti-source Font.plist --interpolate-binary-layout
```

(1) is already supported, (2) is the subject of the current PR, and (3) is the situation that currently requires a two-step approach.

WDYT?

punchcutter:
I definitely agree it's not so great to call fontmake twice. I don't really see (2) being used, but having the option would be good. I did it while exploring ways to deal with a file currently getting prepped for delivery and this was one way to avoid waiting for both https://github.com/googlei18n/glyphsLib/pull/222 and https://github.com/googlei18n/glyphsLib/issues/223.

PR #358, opened 2017-09-28
Add --no-round argument to use designspace roundGeometry=False

moyogo:
See https://github.com/googlei18n/glyphsLib/pull/220

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -87,6 +87,9 @@ def main(args=None):
     outputGroup.add_argument(
         '--family-name',
         help='Family name to use for masters, and to filter output instances')
+    outputGroup.add_argument(
+        '--no-round', dest='no_round', action='store_true',
+        help='Do not round to integers in output when interpolating.')
 
     contourGroup = parser.add_argument_group(title='Handling of contours')
     contourGroup.add_argument(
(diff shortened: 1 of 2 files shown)


Lib/fontmake/font_project.py
```diff
@@ -413,12 +417,14 @@ def run_from_designspace(
 
         ufos = []
         if not interpolate or masters_as_instances:
-            reader = DesignSpaceDocumentReader(designspace_path, ufoVersion=3)
+            reader = DesignSpaceDocumentReader(designspace_path, ufoVersion=3,
+                                               round_geometry=round_geometry)
```
anthrotype: I believe here passing `round_geometry` argument doesn't make any difference, as no interpolation is happening in this branch.

moyogo: Good point.

moyogo:
@anthrotype ping

Lib/fontmake/__main__.py
```diff
@@ -87,6 +87,9 @@ def main(args=None):
     outputGroup.add_argument(
         '--family-name',
         help='Family name to use for masters, and to filter output instances')
+    outputGroup.add_argument(
+        '--no-round', dest='no_round', action='store_true',
+        help='Do not round to integers in ouput when interpolating.')
```
anthrotype: `s/ouput/output/`

PR #365, opened 2017-10-18
Use makeotf.cmd instead of makeotf on win32

moyogo:
On Windows, running the command `makeotf` with subprocess.check_output fails with `FileNotFoundError: [WinError 2] The system cannot find the file specified`.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -586,6 +586,7 @@ class FDKFeatureCompiler(FeatureCompiler):
         if not self.features.strip():
             return
 
+        import sys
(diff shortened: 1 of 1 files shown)


anthrotype:
Hm. Does it work with shell=True? “.cmd” should be in the PATHEXT and cmd.exe normally ignores those

moyogo:
@anthrotype Ah, indeed `shell=True` works.

anthrotype:
Better without shell=True, like you did

https://stackoverflow.com/questions/4965175/make-subprocess-find-git-executable-on-windows

PR #410, opened 2018-03-29
Allow to build selected instances

anthrotype:
add a new `-I/--instance-name` command line option that can be used to select a specific instance (or instances) to generate from a designspace file.

Also uses the new `glyphsLib.to_designpace` API, and depends on https://github.com/googlei18n/glyphsLib/pull/339 for applying instance data on the selected instances.

will add some tests soon

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -16,6 +16,7 @@
 from argparse import ArgumentParser, ArgumentTypeError
 from fontmake import __version__
 from fontmake.font_project import FontProject
+from fontmake.errors import FontmakeError
 
 
 class PyClassType(object):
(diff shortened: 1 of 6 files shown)


anthrotype:
the idea is, e.g. one could call multiple fontmake processes from one's makefile each building a specific instance in parallel.
Or if one wants to, say, build only the "Regular" instances from a bunch of Noto glyphs sources.

I also added a `--designspace-path` option to tell fontmake where to save the designspace file generated from glyphsLib, and a `--master-dir` and `--instance-dir` options to control where the generated UFOs will be created, so we don't hardcode the "master_ufo" and "instance_ufo" directories and users should be free to arrange these as they wish (will do something similar for the output binary formats, probably in a different PR).

anthrotype:
gonna merge this to master, but wait to tag a fontmake release until the new glyphsLib has been released first

PR #412, opened 2018-03-30
WIP Skip decomposing nonexistent components

belluzj:
but now it's fonttools ttLib that fails

Maybe I should erase them from the UFO somehow?


anthrotype:
maybe you should fix it in here:
https://github.com/googlei18n/ufo2ft/blob/master/Lib/ufo2ft/filters/decomposeComponents.py

i want to make fontmake use these ufo2ft filters instead of doing its own decomposing here.

I think in the ufo2ft decomposing filter  we should add empty stub glyphs for these missing components, so the ttGlyphPen will complete.

PR #422, opened 2018-04-25
add --output-path and --output-dir options

anthrotype:
The two options are mutually exclusive.

`--output-path` can only be used when the output is a single file: e.g. when the input is a single UFO, or when only one instance was selected for export, or the source contains only one master, or the output format is a variable font.

The `--output-dir` option can be used with any binary output. The folder is created if it doesn't exist.

Finally, the `--interpolate-binary-layout` option now takes an optional parameter to customize the location of the binary masters used with `varLib.interpolate_layout` (e19f00f).

Fixes https://github.com/googlei18n/fontmake/issues/333 and https://github.com/googlei18n/fontmake/issues/396

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -226,7 +226,8 @@ def main(args=None):
         project.run_from_ufos(
             ufo_paths, is_instance=args.pop('masters_as_instances'), **args)
     except FontmakeError as e:
-        parser.error(e)
+        import sys
+        sys.exit("fontmake: error: %s" % e)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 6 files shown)


anthrotype:
but instances in designspace files already have a filename attribute set, that's how the format is supposed to work. You need to set them accordingly in your designspace document. When building from glyphs source, fontmake will set the instance directory in the generated designspace document. But when the user provides a designspace document, it simply does what it's asked, i.e. generate the instances with the given filenames.

madig:
Hm. Then we need to fix glyphsLib to make sure the instance filenames are different or put in a different directory.

Optionally: make --instance-dir work for designspace files too if the filename is relative and complain if it's absolute. It's kind of like DESTDIR.

PR #423, opened 2018-04-25
optionally save master/instance ufos to temporary directory

anthrotype:
In https://github.com/googlei18n/fontmake/pull/410#issuecomment-377329203, I added --master-dir and --instance-dir options to control where the glyphsLib-generated UFOs are created, and override the hardcoded "master_ufo"/"instance_ufo" directories.

Sometimes one may only be interested in running fontmake end-to-end from .glyphs sources to binary TTF or OTFs (or variable TTF), and thus the master_ufo and instance_ufo folders that are created in the current working directory are just unwanted leftovers.

This PR allows to pass to `--master-dir` or `--instance-dir` a special string `{tmp}`, which means create a temporary directory and use that, then clean it up automatically after finished.

I chose `{tmp}` because is short and it's unlikely that one wants to use it as the real name for an existing filesystem folder (giving the full path should work for those edge cases).

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 
-import shutil
+from contextlib import contextmanager
 from argparse import ArgumentParser, ArgumentTypeError
 from fontmake import __version__
 from fontmake.font_project import FontProject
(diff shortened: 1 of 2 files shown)


anthrotype:
I know.. I wanted to reduce the amount of surprise for users who may be used to the current way.
If others say this is ok, we change this to be the default behaviour.

moyogo:
`--no-master-dir` and `--no-instance-dir` may be clearer.

PR #440, opened 2018-06-06
update to use current ufo2ft API

anthrotype:
decomposing glyphs, removing overlaps, converting curves to quadratic can be done by appropriate ufo2ft preprocessor filters, fontmake only needs to pass the right options to the top-level compile functions.
Note that this was already implemented in the current stable ufo2ft 1.1.0 released in december last year, we just didn't update fontmake to make use of them.

Also, adds support for `cff_round_tolerance`. In ufo2ft the argument is called simply `roundTolerance` but here we need to distinguish it from the other one called `round_instances`.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -20,7 +20,7 @@ import glob
 import logging
 import math
 import os
-from functools import partial
+from functools import partial, wraps
 import tempfile
 import shutil
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


Lib/fontmake/font_project.py
```diff
@@ -306,25 +349,35 @@ def save_otfs(
...
+                **compiler_options)
+
         do_autohint = ttf and autohint is not None
-        for ufo in ufos:
-            name = self._font_name(ufo)
-            logger.info('Saving %s for %s' % (ext.upper(), name))
-
-            if use_production_names is None:
-                use_production_names = not ufo.lib.get(
-                    GLYPHS_PREFIX + "Don't use Production Names")
```
anthrotype: This Glyphs.app-specific UFO lib key will also be supported in the upcoming ufo2ft
https://github.com/googlei18n/ufo2ft/commit/2ca438c0e7779b092f44f204d7bda666bb9af68b

PR #444, opened 2018-06-14
subset_otf_from_ufo: Handle .notdef glyph

madig:
The glyph names of UFO anf OT binary need to match up for the export flag testing to work. The .notdef glyph must be handled specially because it is usually the first glyph in the OT, but might not be present or in the first glyph slot in the UFO.

https://github.com/googlei18n/fontmake/issues/443

Now for some tests.

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -431,7 +431,13 @@ class FontProject(object):
                 os.remove(otf_path)
 
     def subset_otf_from_ufo(self, otf_path, ufo):
-        """Subset a font using export flags set by glyphsLib."""
+        """Subset a font using export flags set by glyphsLib.
+
+        There are two more settings that can change export behavior:
+        "Export Glyphs" and "Remove Glyphs", which are currently not supported
+        for complexity reasons. See
+        https://github.com/googlei18n/glyphsLib/issues/295.
+        """
         ufo_order = makeOfficialGlyphOrder(ufo)
         ot_order = TTFont(otf_path).getGlyphOrder()
         assert ot_order[0] == ".notdef"
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -51,6 +51,7 @@ from fontTools import designspaceLib
 from fontTools.varLib.interpolate_layout import interpolate_layout
 from ufo2ft import compileOTF, compileTTF, compileInterpolatableTTFs
 from ufo2ft.featureCompiler import FeatureCompiler
+from ufo2ft.util import makeOfficialGlyphOrder
 
 from fontmake.errors import FontmakeError, TTFAError
 from fontmake.ttfautohint import ttfautohint
@@ -435,20 +436,17 @@ class FontProject(object):
         keep_glyphs = set(ufo.lib.get(GLYPHS_PREFIX + 'Keep Glyphs', []))
 
         include = []
-        ufo_order = [glyph_name
-                     for glyph_name in ufo.lib[PUBLIC_PREFIX + 'glyphOrder']
-                     if glyph_name in ufo]
+        ufo_order = makeOfficialGlyphOrder(ufo)
         ot_order = TTFont(otf_path).getGlyphOrder()
-
-        # An OpenType binary usually has the .notdef glyph as the first glyph,
-        # which might or might not be present in the UFO, and not necessarily as
-        # the first glyph. Manipulate the order to make glyph names match up in
-        # the loop below.
-        if ".notdef" in ufo:
-            ufo_order.pop(ufo_order.index(".notdef"))
-            ufo_order.insert(0, ".notdef")
-        else:
-            ot_order = ot_order[1:]  # Strip out ".notdef"
+        assert ot_order[0] == ".notdef", (
+            "{}, subsetting: .notdef must be the first glyph in "
+            "{}.".format(ufo.path, otf_path)
+        )
+        assert len(ufo_order) == len(
+            ot_order
+        ), "{}, subsetting: amount of glyphs does not match with {}".format(
+            ufo.path, otf_path
+        )
 
         for old_name, new_name in zip(ufo_order, ot_order):
             glyph = ufo[old_name]
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -438,15 +438,8 @@ class FontProject(object):
         include = []
         ufo_order = makeOfficialGlyphOrder(ufo)
         ot_order = TTFont(otf_path).getGlyphOrder()
-        assert ot_order[0] == ".notdef", (
-            "{}, subsetting: .notdef must be the first glyph in "
-            "{}.".format(ufo.path, otf_path)
-        )
-        assert len(ufo_order) == len(
-            ot_order
-        ), "{}, subsetting: amount of glyphs does not match with {}".format(
-            ufo.path, otf_path
-        )
+        assert ot_order[0] == ".notdef"
+        assert len(ufo_order) == len(ot_order)
 
         for old_name, new_name in zip(ufo_order, ot_order):
             glyph = ufo[old_name]
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -450,10 +450,10 @@ class FontProject(object):
 
         for old_name, new_name in zip(ufo_order, ot_order):
             glyph = ufo[old_name]
-            if ((keep_glyphs and old_name not in keep_glyphs) or
-                not glyph.lib.get(GLYPHS_PREFIX + 'Glyphs.Export', True)):
-                continue
-            include.append(new_name)
+            in_keep_glyphs = keep_glyphs and old_name not in keep_glyphs
+            exported = glyph.lib.get(GLYPHS_PREFIX + "Glyphs.Export", True)
+            if in_keep_glyphs or exported:
+                include.append(new_name)
 
         # copied from nototools.subset
         opt = subset.Options()
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -438,9 +438,19 @@ class FontProject(object):
         ufo_order = [glyph_name
                      for glyph_name in ufo.lib[PUBLIC_PREFIX + 'glyphOrder']
                      if glyph_name in ufo]
-        for old_name, new_name in zip(
-                ufo_order,
-                TTFont(otf_path).getGlyphOrder()):
+        ot_order = TTFont(otf_path).getGlyphOrder()
+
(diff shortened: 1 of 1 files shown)


madig:
Hum. There doesn't seem to be a dedicated place for testing this. Well, it Works For Me(tm).

moyogo:
We should probably us `ufo2ft.util.makeOfficialGlyphOrder`, ufo_order isn’t correct if there are glyphs that are not in the font.glyphOrder.

Lib/fontmake/font_project.py
```diff
@@ -438,9 +438,19 @@ def subset_otf_from_ufo(self, otf_path, ufo):
...
-                TTFont(otf_path).getGlyphOrder()):
+        ot_order = TTFont(otf_path).getGlyphOrder()
+
+        # An OpenType binary usually has the .notdef glyph as the first glyph,
+        # which might or might not be present in the UFO, and not necessarily as
+        # the first glyph. Manipulate the order to make glyph names match up in
+        # the loop below.
+        if ".notdef" in ufo:
+            ufo_order.pop(ufo_order.index(".notdef"))
+            ufo_order.insert(0, ".notdef")
```
anthrotype: what about `ufo_order.insert(0, ufo_order.pop(ufo_order.index(".notdef")))` ?

Lib/fontmake/font_project.py
```diff
@@ -438,9 +438,19 @@ def subset_otf_from_ufo(self, otf_path, ufo):
...
+        else:
+            ot_order = ot_order[1:]  # Strip out ".notdef"
```
anthrotype: maybe you should check (or assert) that .notdef is really the first glyph in the TTFont. I mean, we know it is, because ufo2ft, but checking doesn't harm

Lib/fontmake/font_project.py
```diff
@@ -450,10 +450,10 @@ def subset_otf_from_ufo(self, otf_path, ufo):
...
         for old_name, new_name in zip(ufo_order, ot_order):
             glyph = ufo[old_name]
-            if ((keep_glyphs and old_name not in keep_glyphs) or
-                not glyph.lib.get(GLYPHS_PREFIX + 'Glyphs.Export', True)):
-                continue
-            include.append(new_name)
+            in_keep_glyphs = keep_glyphs and old_name not in keep_glyphs
+            exported = glyph.lib.get(GLYPHS_PREFIX + "Glyphs.Export", True)
+            if in_keep_glyphs or exported:
+                include.append(new_name)
```
anthrotype: oh, since you're touching this, I think you should have a look at this conversation with @schriftgestalt
https://github.com/googlei18n/fontmake/issues/381
and 
https://github.com/googlei18n/glyphsLib/issues/295

I forgot the details about it and don't have time to investigate now

Lib/fontmake/font_project.py
```diff
@@ -435,17 +436,24 @@ def subset_otf_from_ufo(self, otf_path, ufo):
...
-                     if glyph_name in ufo]
-        for old_name, new_name in zip(
-                ufo_order,
-                TTFont(otf_path).getGlyphOrder()):
+        ufo_order = makeOfficialGlyphOrder(ufo)
+        ot_order = TTFont(otf_path).getGlyphOrder()
+        assert ot_order[0] == ".notdef", (
+            "{}, subsetting: .notdef must be the first glyph in "
+            "{}.".format(ufo.path, otf_path)
+        )
```
anthrotype: I see you like it black :black_heart: 
I guess you `git add -p`, since black always reformat the whole file.
in another pr, we should blacken it all and forget about it

Lib/fontmake/font_project.py
```diff
@@ -435,17 +436,24 @@ def subset_otf_from_ufo(self, otf_path, ufo):
...
-                     for glyph_name in ufo.lib[PUBLIC_PREFIX + 'glyphOrder']
-                     if glyph_name in ufo]
-        for old_name, new_name in zip(
-                ufo_order,
-                TTFont(otf_path).getGlyphOrder()):
+        ufo_order = makeOfficialGlyphOrder(ufo)
+        ot_order = TTFont(otf_path).getGlyphOrder()
+        assert ot_order[0] == ".notdef", (
+            "{}, subsetting: .notdef must be the first glyph in "
+            "{}.".format(ufo.path, otf_path)
```
anthrotype: this assert is for the devs' own consumption, not the user's. I think the error message is too much in this case

Lib/fontmake/font_project.py
madig: I see you have good 👀

Lib/fontmake/font_project.py
```diff
@@ -432,21 +432,29 @@ def save_otfs(self,
...
+            keep_glyphs = set(keep_glyphs_list)
+        else:
+            keep_glyphs = None
+
+        include = []
+        for source_name, binary_name in zip(ufo_order, ot_order):
+            glyph = ufo[source_name]
+
+            if keep_glyphs and source_name not in keep_glyphs:
+                continue
```
anthrotype: ok, so now Keep Glyphs will take priority over individual glyph export flags. Did you check Glyphs.app works like that?

madig:
Add summary where? Here?

Glyphs' export flag on single glyphs and the global "Keep Glyphs" setting are now correctly handled when subsetting is implicitly (by setting the flags) or explicitly requested.

madig:
Yes

anthrotype:
Thanks

anthrotype:
a byproduct of this PR is that the post-build subsetting feature no longer requires that the UFO explicitly defines `public.glyphOrder` in lib.plist, since we are now using the same function ufo2ft uses to make a glyph order, and this always returns something even when glyphOrder is not defined.

chrissimpkins:
@anthrotype @madig I think that this impacts us.  I am maintaining a separate subset lib.plist file by hand for our subset definitions.  How do we specify the subset Unicode code points following this change?   If this is a big refactor on the definition file end, it might be the push necessary for me to move to a different approach.  The current one is crazy but works so I haven't addressed it.  Also, what are your thoughts about supporting a non-glyphs library (and maybe even external to UFO spec) specific way to support subset definitions in fontmake.  Perhaps an approach from command line like pyftsubset or a json/yml definition file down the road?

anthrotype:
there is no change, it continues to work like before. What changed is the situation where one has _both_ the global "Keep Glyphs" list, and the individual glyph Export flags. The former global list always wins. The latter are used only then the former is not defined. As for the mismatching glyphOrder between the UFO and the TTFont, I don't think that should impact you.

anthrotype:
> a non-glyphs library (and maybe even external to UFO spec) specific way to support subset definitions in fontmake

see https://github.com/unified-font-object/ufo-spec/issues/68

anthrotype:
Note that the reason we use the fonttools subsetter ex-post, instead of subsetting the source UFOs (in ufo2ft, for example), is that it's trivial to remove glyphs from a font, its kerning or groups, but subsetting the feature file is not straightforward. Granted, we could remove all reference to a dropped glyph in all the feaLib AST statements, and drop the empty statements, blocks, etc. ... but calling the fonttools subsetter is too convenient.

anthrotype:
related https://github.com/fonttools/fonttools/issues/1121#issuecomment-350015932

chrissimpkins:
> there is no change, it continues to work like before.

I am reading this statement as it is ok to specify both `<key>public.glyphOrder</key>` with the full set list and `<key>com.schriftgestaltung.Keep Glyphs</key>` with only the subset glyph definitions in the same lib.plist file and behavior does not change.  You still get only the code points specified in `com.schriftgestaltung.Keep Glyphs` (and ignore public.glyphOrder definitions for the Unicode positions included in the subsets irrespective of ordering) with these changes?  Sorry, I am making sure that I understand your comment above.

anthrotype:
>  it is ok to specify both <key>public.glyphOrder</key> with the full set list and <key>com.schriftgestaltung.Keep Glyphs</key> with only the subset glyph definitions 

totally, yes! sorry if I confused you with my previous statement.

it's not "codepoints" technically, but glyph names, but yeah :+1:

chrissimpkins:
>  but calling the fonttools subsetter is too convenient.

That was not intended to be a criticism of the implementation approach, rather wearing the compiler user cap it would be nice to be able to specify definitions, including specification of multiple subset definitions simultaneously if it is to be part of a source spec, through different means.   I commented in the ufoLib thread with some ideas about this that would be very helpful to us.  We would like to create a number of subset files and I can't support this with our current approach.

https://github.com/unified-font-object/ufo-spec/issues/68#issuecomment-397331040

chrissimpkins:
Thanks Cosimo.  Appreciate your help and feedback.

Lib/fontmake/font_project.py
```diff
@@ -430,22 +431,36 @@ def save_otfs(self,
...
-        keep_glyphs = set(ufo.lib.get(GLYPHS_PREFIX + 'Keep Glyphs', []))
+        There are two more settings that can change export behavior:
+        "Export Glyphs" and "Remove Glyphs", which are currently not supported
+        for complexity reasons. See
+        https://github.com/googlei18n/glyphsLib/issues/295.
+        """
+        ufo_order = makeOfficialGlyphOrder(ufo)
+        ot_order = TTFont(otf_path).getGlyphOrder()
+        assert ot_order[0] == ".notdef"
+        assert len(ufo_order) == len(ot_order)
```
anthrotype: ouch. this fails if the UFO does not originally contain the ".notdef" glyph. Ufo2ft always insert one when missing, but that `makeOfficialGlyphOrder(ufo)` function does not add ".notdef" if it's not already in there...
only if it's present, it forces it to be the first, otherwise it ignores it.
https://github.com/googlei18n/ufo2ft/blob/6b710f7e889368746cbcf4340cbf994998cea1ae/Lib/ufo2ft/util.py#L36-L39

in the ufo2ft.outlineCompiler, that function is called on the resulting glyphSet only after the makeMissingRequiredGlyphs method has inserted a .notdef.
But here we are calling it on the input UFO.

anthrotype: We can check if '.notdef' is not in ufo, then add one at the beginning of the ufo_order list..
This could work for the time being, since the only "missing required" glyph that is being added by ufo2ft is the .notdef, but it's possible that we add more later. Relying on that implementation detail is not very future-proof.
🤔

madig: Ouch!

Shouldn't `makeOfficialGlyphOrder` really return the official one, with all required glyphs in it? Otherwise, maybe `makeMissingRequiredGlyphs` can be refactored into `missing_required_glyphs` and `insert_missing_required_glyphs` or somesuch? Then `makeOfficialGlyphOrder` can call the former.

PR #459, opened 2018-09-06
Fixes #458

### LINKED ISSUE #458: -M option (masters_as_instances) crashes
Multiple issues on this code path, especially in `Lib/fontmake/font_project.py`:

* the current code in run_from_ufos doesn’t sanitize its input, if the -M option is passed from a designspace, the list of ufos fed into run_from_ufos is a spurious list of paths and defcon.Font objects
* the FontProject._font_name private method doesn’t take into account that a UFO might not have an empty familyName or styleName—I edited the code to ameliorate this, but ufo2ft really didn’t like a UFO like that. I ended up adding familyName and styleName entries to the fontinfo.plist in MyTest-Light.ufo and MyTest-Bold.ufo in the AvarDesignspace tests.

fontmake’s test are somewhat lacking at the moment, however I did add a test to test/run.sh to try and make sure this functionality would at least be excercised by the test harness. 

I have a fix on hand.

verbosus:


Lib/fontmake/font_project.py
```diff
@@ -650,14 +650,19 @@ def run_from_ufos(
...
+
+        # the `ufos` parameter can be a list of UFO objects
+        # or it can be a path (string) with a glob syntax
+        ufo_paths = []
+        if isinstance(ufos, basestring):
+            ufo_paths = glob.glob(ufos)
+            ufos = [Font(x) for x in ufo_paths]
+        elif isinstance(ufos, list):
+            # ufos can be either paths or open Font objects, so normalize them
+            ufos = [Font(x) if type(x) is not Font else x for x in ufos]
```
anthrotype: Please use isinstance here instead of type()

anthrotype: Nevermind i’ll do the fixup

PR #484, opened 2018-11-29
Expand master feature file to instance UFOs

madig:
I want to tackle https://github.com/googlei18n/fontmake/issues/359.

Rather than changing the paths (complicated because you have to rewrite the AST recursively), here I just expand the feature file to every instance. Not sure if anyone needs instance UFOs for anything other than generating binaries?

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -653,7 +653,7 @@ class FontProject(object):
             ufos.extend(apply_instance_data(designspace_path,
                                             include_filenames=filenames))
             if expand_features_to_instances:
-                logger.info("Expanding features to instance UFOs")
+                logger.debug("Expanding features to instance UFOs")
                 master_source = next(
                     (s for s in designspace.sources if s.copyFeatures), None
                 )


Lib/fontmake/font_project.py
```diff
@@ -644,6 +644,18 @@ def run_from_designspace(
             logger.info('Applying instance data from designspace')
             ufos.extend(apply_instance_data(designspace_path,
                                             include_filenames=filenames))
+            logger.info("Expanding features to instance UFOs")
+            master_source = next(
+                (s for s in designspace.sources if s.copyFeatures), None
+            )
+            if not master_source:
+                logger.info("... actually, no source is marked as the feature master")
```
anthrotype: Let’s avoid too verbose logging messages. Either demote to DEBUG or remove completely

Lib/fontmake/font_project.py
```diff
@@ -644,6 +644,18 @@ def run_from_designspace(
             logger.info('Applying instance data from designspace')
             ufos.extend(apply_instance_data(designspace_path,
                                             include_filenames=filenames))
+            logger.info("Expanding features to instance UFOs")
```
anthrotype: Maybe we should only expand if there is any include statement, or put this behind an option?

madig: I'll turn it into an option.

madig: Dunno. Can change to opportunistically expanding stuff unless someone complains?

madig: @miguelsousa would autoexpansion impact you? Adobe is the only other big UFO user I know.

miguelsousa: What is autoexpansion?

madig: The goal is work around the problem of relative feature file `include()` paths that get invalidated when fontmake generates instance UFOs for further processing and puts these somewhere else. One solution is  to manually copy stuff over, the other is parsing the feature file of the source marked with `<features copy="1" />`, resolving all includes and writing the fully expanded feature file to the instance UFOs, which would then be independent of external files: autoexpansion.

madig: Out of curiosity, how would it impact you and what is the actual workflow you're using?

miguelsousa: Look at the Source Sans Pro project

anthrotype:
@madig what's the status of this?
It's marked as WIP but LGTM

madig:
miguelsousa says it would affect their workflow negatively, so, this option can stay optional for now I guess? Maybe it should really be part of ufoProcessor when fontmake switches to that.

PR #512, opened 2019-02-04
Housekeeping

madig:


.editorconfig
```diff
@@ -0,0 +1,15 @@
+# http://editorconfig.org
```
anthrotype: we don't need this if we are going to autoformat with black

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  rev: v2.1.0
+  hooks:
+  - id: trailing-whitespace
+  - id: end-of-file-fixer
+  - id: check-yaml
+  - id: check-ast
+  - id: mixed-line-ending
+    args: ["--fix=lf"]
+  - id: debug-statements
+  - id: flake8
```
anthrotype: flake8 official repo now exports its own pre-commit hook, e.g. see https://github.com/asottile/pyupgrade/pull/95

.editorconfig
madig: It helps VSCode do the right thing. I tried.

.pre-commit-config.yaml
madig: Oh nice. Can't find an immediate mention of bugbear but it doesn't seem to complain about the source.

Issue #359, opened 2017-09-29
Expand included features file when generating UFOs


moyogo:
When a source master UFO has a features.fea file with an `include` statement, it should be expanded into that file.
We can’t just rewrite the `include` statement with the relative path since the FDK semantics won’t allow that for more than one level deep.

anthrotype:
could you elaborate a bit more, or provide an example to understand better what is the problem?
also, would you like to work on that?

moyogo:
See https://travis-ci.org/moyogo/fontmake/jobs/281372224 with the branch [included_features](https://github.com/googlei18n/fontmake/compare/master...moyogo:included_features?expand=1).

anthrotype:
I see. So every time there's an include with a relative path a master UFO (or even in a *.glyphs source, I guess), we can't simply copy the feature text from the master to the instances, because by default we are saving the instances in a different sub-folder than the masters, so the relative path does not work any more. 
Perhaps, instead of expanding the include statements and flattening the feature text, fontmake could replace the relative paths with the correct ones. So in your example, instead of `../family.fea` (i.e. relative to the master UFO's including features.fea), it would write `../../family.fea` (relative to the instance UFO's, which is created in ./instance_ufos sub-folder).

madig:
Can be done with `--expand-features-to-instances`. Should maybe be the default one day.

anthrotype:
@moyogo wrote in https://github.com/googlefonts/fontmake/pull/484

> Do we really need an option for this? What’s the use case for wanting broken instance UFOs with include statements pointing to nowhere?

I agree with him, it should not be necessary for the user to pass an option to build valid instances. If someone has peculiar workflows that require _not_ to expand the features to the static instances, we can add an option to disable this. Let's flip the default value of --expand-features-to-instances to be True and add the negative option.

anthrotype:
actually, I think the cleanest solution would be to walk the feature file's AST and adjust the relative paths of the `include()` statements accordingly.
We would need to do the same, not only when master UFOs contain include(), but also if .glyphs source files contain them, as we generate master UFOs from the latter inside a different folder, thus breaking relative paths in include().
Maybe fontTools feaLib should expose an API to update the paths of included files.

anthrotype:
related glyphsLib issue https://github.com/googlefonts/glyphsLib/issues/506

PR #513, opened 2019-02-04
Optionally expand feature files to instance UFOs

madig:
I want to tackle #359.

Rather than changing the paths (complicated because you have to rewrite the AST recursively), here I just expand the feature file to every instance. Optional because Adobe's workflow apparently is to manage this manually.


madig:
https://travis-ci.org/googlei18n/fontmake/jobs/488636575#L691

Meh.

anthrotype:
The -i option doesn’t make sense with -o ttf-interpolatable. The former makes static instances (which need not have interpolatable outlines), the latter is only required to make variable fonts. I made the two mutually exclusive to simplify the rest of the code.

PR #518, opened 2019-02-05
Bump requirements for fontTools

madig:
We need 3.37.1 for the MVAR generation fix.


madig:
Great. Works on Py2, crashes on Py3...

madig:
Tracked down to https://github.com/fonttools/fonttools/commit/ef1d4cd02d1e46f5dac3914f547a6e4275cf3077. Not sure what to do.

anthrotype:
Check in case there are circular references in cff table back to the TTFont it’s contained in..

PR #521, opened 2019-02-07
Improve installation instructions

madig:
https://github.com/googlei18n/fontmake/issues/520

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,10 +13,47 @@ Installation
 
 Fontmake requires Python 2.7, 3.5 or later.
 
-Releases are available on `PyPI`_ and can be installed with `pip`_.
+There are three ways of getting fontmake.
+
+Installing from PyPI
+--------------------
+
+The standard route recommended for most people. Releases are available on
+`PyPI`_ and can be installed with `pip`_. In these instructions, we install
+fontmake in a virtual environment to avoid software clashes and permission
+problems. Open a terminal and follow along:
 
 .. code:: bash
 
+    # This assumes you have Python installed. If you are unsure, get Python 3
+    # from www.python.org.
+
+    # First, go to your project folder.
+
+    # Then, create a new virtual environment. You only need to do this the first
+    # time or if you deleted the old one.
+    # If you are using Python 2:
+    python -m virtualenv venv
+    # Otherwise, if you are using Python 3 on Mac or Linux:
+    python3 -m venv venv
+    # Otherwise, if you are using Python 3 on Windows:
+    py -3 -m venv venv
+
+    # Note that if you use Git to manage your project, you should add the line
+    # "venv" to the text file ".gitignore". Create the file in the top directory
+    # of your project if it doesn't exist. This will prevent accidentally
+    # committing the entire venv to Git, which is most probably not what you want.
+
+    # Activate the virtual environment. You need to do this every time you open
+    # a new terminal.
+    # On Mac or Linux, with the standard Bash shell:
+    . venv/bin/activate
+    # On Windows, with a cmd.exe terminal
+    .\venv\Scripts\activate.bat
+    # On Windows, in Powershell:
+    .\venv\Scripts\Activate.ps1
+
+    # Install fontmake.
     pip install fontmake
 
 Use the ``-U``, ``--upgrade`` option to update fontmake and its dependencies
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -13,42 +13,98 @@ Installation
...
+If you have Python 3.6 or later installed, you can get a zip file with a
+stand-alone fontmake application from the [GitHub release page]_.
+
+1. Pick one for your platform and Python version, a "32" in the name needs a
+   32-bit Python installation, a "64" a 64-bit one. If you are unsure, better
+   follow the PyPI route above instead.
+2. Unzip it to where you want it, e.g. your project folder.
+3. Open a terminal and go to that folder.
+4. On Windows, run ``python fontmake.pyz``, on other platforms run
+   ``python3 fontmake.pyz``. If you see a message telling you about the options
```
anthrotype: actually this is incorrect. The way you run the shiv bundled app is simply with its name (provided it is in the `PATH`), no need to call it with a specific python interpreter. On Windows, the `py` launcher will read the shebang and call the pyz app with the right interpreter. On Unix, bash or other shell will do the right thing (as long as `python3.6` or `python3.7` -- depending on which `fontmake.pyz` was downloaded -- are in the `PATH`).

anthrotype: On Windows, the `.pyz` extension is normally registered to be executed by the `py` python launcher which is installed by default with Python.org installers. If that is not the case for whatever reason, then simpy running `fontmake.pyz` may not work out of the box, and you may need to either associated `.pyz` extesion with the python launcher, or call it as `py fontmake.pyz`

anthrotype:
maybe we should recommend using [pipx](https://github.com/pipxproject/pipx) to make it easier to install/upgrade without needing to use virtual envs

MalcolmWooden:
Currently I tell people in our team to use this command for updating:
"python3 -m pip install -U fontmake"

Will the pipx installer do the same thing (without walking over other installations)?

anthrotype:
I haven't tried it yet, but I heard it's gaining momentum in the python community and was designed to make the life simpler for people who want to install (and upgrade) python applications that are hosted on PyPI, without having to deal with the complications of virtual environments.
Maybe tried it and let us know if it works for you?
Fontmake should already work right now if installed via this pipx installer.

chrissimpkins:
Maybe Pyinstaller builds to avoid the entire issue?  Would need to build by platform on platform but could likely do this through CI based releases.

anthrotype:
I just tried pipx on my machine and it's really cool. I think i'm going to use this from now on, for python CLI tools that I simply use and am not an active developer of.

pipx gives you the isolation of virtual environments without the hassle of needing to activate them before use, and without having to know how to modify your shell configuration to ensure the `~/.local/bin` folder is in the PATH (it has a command to do that automatically, that's handy for beginners).

The pyinstaller approach has the only advantage that it doesn't require python to be installed, but most of us font developers have python installed anyway, one way or another.

anthrotype:
so the instructions for fontmake with pipx would become simply:

```
$ python3 -m pip install --user pipx
$ python3 -m userpath append ~/.local/bin
$ pipx install fontmake  # or pipx upgrade fontmake
$ fontmake --help
```

anthrotype:
another advantage of pipx over pyinstaller and shiv, is that it allows to automatically upgrade the application, whereas for the latter two approaches one has to download the updated standalone app (curl, wget, or manually browse and click download link), unzip it and move it to the desired location on the $PATH. With pipx it's simply `pipx upgrade fontmake`

chrissimpkins:
re pyinstaller it also pins the Python interpreter version within the build without the end user needing to worry what version they are on or attempting to install in Py interpreter specific environment.  mv to system PATH and off you go.

chrissimpkins:
>another advantage of pipx over pyinstaller and shiv, is that it allows to automatically upgrade the application, whereas for the latter two approaches one has to download the updated standalone app (curl, wget, or manually browse and click download link), unzip it and move it to the desired location on the $PATH. With pipx it's simply `pipx upgrade fontmake`

good point

anthrotype:
oh, and there's also `pipx upgrade-all` which upgrades all applications installed via pipx in one go!

madig:
Just tried it. I like it. Ok let's recommend that :)

Issue #526, opened 2019-02-17
Input arguments don't include a Glyphs app generated VF


EbenSorkin:
The instructions as of now are 

Input arguments:
  The following arguments are mutually exclusive.

  -g GLYPHS, --glyphs-path GLYPHS
                        Path to .glyphs source file
  -u UFO [UFO ...], --ufo-paths UFO [UFO ...]
                        One or more paths to UFO files
  -m DESIGNSPACE, --mm-designspace DESIGNSPACE
                        Path to .designspace file

I would like to be able to process a Glyphs app variable font. May I use the  -g or is this in fact not supported?

Side question - it looks like Glyphs outputs a TTF flavored VF only. Is that correct?

davelab6:
G should work!

Yes afaik TTF only, fdk for cff2, building from ufo

EbenSorkin:
Thanks!

-e.

madig:
`fontmake -g myfont.glyphs -o variable`

Currently, only TTF VFs are supported by fontTools. CFF2 support is pending.

PR #528, opened 2019-02-17
Lib/fontmake/__main__.py tweak help strings

davelab6:
Prevent user confusion like #526 

Please squash merge


madig:
Thanks :)

I think the confusion came from fontmake not following the usual conventions of taking loose arguments as input/output paths, but maybe it helps.

PR #533, opened 2019-02-27
Add --no-varlib-optimize switch

madig:
Tackle https://github.com/googlei18n/fontmake/issues/532.

Doesn't work currently because the code in `main` deleting arguments depending on arguments is too clever for my brain right now.


anthrotype:
will be superseded by #551

PR #554, opened 2019-05-18
Rework testsuite

madig:


madig:
I marked one test as skip because I don't know whether the code is broken or I'm doing it wrong. I think the other test directory can go now?

madig:
Thought: leave this Python 3 only and merge to a new branch dev-py3 instead of master, where the rest of the development for a Python 3 release can go.

madig:
Another thought: clean up test data? The various `*DesignspaceTest` folders could maybe be consolidated.

madig:
Note: Use https://github.com/kvas-it/pytest-console-scripts.

anthrotype:
> Note: Use https://github.com/kvas-it/pytest-console-scripts.

Actually about that, I ended up not using it anymore because I don’t like that it expects the scripts to be on $PATH. I don’t like that. I prefer that only the lib in on the PYTHONPATH for the tests to work. Also running subprocesses slows the tests too much.
If you are testing code within main(), keep calling that directly within the current process but monkeypatch  the logging configuration so it has no side effects on the other tests.
Otherwise make things modular and only unit tests those.

madig:
Blargh... tox would probably come in handy...

madig:
Rip out shiv making and the associated `after_success:` Travis step?

anthrotype:
> Rip out shiv making and the associated after_success: Travis step?

i'd like the zipapps to stay. Not sure how many ppl use them though, but doesn't cost us much to make and publish them.

PR #555, opened 2019-05-26
New instantiator

madig:
Build on top of the test rework to add an instantiator to (eventually) replace MutatorMath.

Todo:
- [x] Make it user-selectable to switch between MutatorMath and the new instantiator.
- [x] add handling of `slnt` axis (like for `wght` and `wdth`) which translates 1:1 to post.italicAngle, with clamping.
- [x] Check that data copied to instances from the Instantiator object is independent (don't repeat the hours of fun we had with the ghost kerning bug in varLib.merger)
- [x] How to handle glyphs blacklisted by `public.skipExportGlyphs` that do not interpolate? Write out as empty outlines and warn, but don't error out.

Deferred:

- Do we test for rounding of anchor and component coordinates? (I think this is more a test of fontMath, so no)
- More elegant handling of UFO filename generation for the Instantiator code path. Refactor Designspace loading to fill in instance filepaths? Needs access to global computed paths, etc.
- Switch from defcon to ufoLib2 completely
- Decide on what to do about `Glyph.lib`, clear out for now in instances.


behdad:
This is great, thanks!  Why is it in fontmake though, shouldn't it be in ufo2ft or somewhere lower?

madig:
@anthrotype and me want to test drive it in fontmake first (I use it for Cantarell already) and maintaining an external lib is extra admin. Fontmake should be the ideal temporary home, as it is a self-contained app hat should not be used as a library.

The thinking is, we want to merge ufoLib2 into fontTools.ufoLib.objects one day, reimplement fontMath on those objects and then we figure out where to stick this code. I think.

madig:
I need some more tests that test rounding on vs. off, axis mappings and rules. Anything else?

Also:
- Support SourceDescriptor.mutedGlyphNames?
- Support SourceDescriptor.muteKerning?
- What to do with a glyph's lib? I currently drop it because there's no useful interpolating data afaik, but maybe I'm wrong? What about `public.verticalOrigin` and `public.postscript.hints`? fontMath doesn't interpolate that.

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,499 @@
...
+
+It is an alternative to mutatorMath (used internally by fontmake) and ufoProcessor. The
+aim is to be a minimal implementation that is focussed on using ufoLib2 for font data
+abstraction, varLib for instance computation and fontMath as a font data shell for
+instance computation directly and exclusively.
+
+At the time of this writing, varLib lacks support for anisotropic (x, y) locations and
+extrapolation.
+
+See the `scripts/make-static-fonts.py` script for how to use it.
```
anthrotype: the line above is probably a copy-paste leftover

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,499 @@
...
+        font.lib["designspace.location"] = list(location.items())
+
+        return font
+
+
+def anisotropic(location: Location) -> bool:
+    for v in location.values():
+        if isinstance(v, tuple):
+            return True
+    return False
```
anthrotype: replace with `return any(isinstance(v, tuple) for v in locations.values())`

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,499 @@
...
+
+
+def normalize_design_location(
```
anthrotype: this `normalize_design_location` function does nothing more than `varLib.models.normalizeLocation`

anthrotype:
about those "muted" attributes for source kerning, info and glyphs, what are they used for? I don't know how widely used they are in the wild, but varLib doesn't read them when generating varfonts.

anthrotype:
> What about public.verticalOrigin and public.postscript.hints?

maybe once public.verticalOrigin is upgraded to a proper item in the next version of UFO, we can add support for that too
https://github.com/unified-font-object/ufo-spec/issues/95

Not sure about postscript hints. Were they expected to be interpolated in the mutatormath workflow? Can be done later, I guess

madig:
(Idea to self: fill in PANOSE weight field from wght axis maybe)

madig:
Regarding "muted" attributes, shall we remove them to match what varLib does? Makes sense from that standpoint. Plus, they can be added back in I guess.

madig:
I somehow managed to kill instantiator.py while rebasing. Hope I didn't loose anything.

madig:
I am wondering: `swapGlyphNames` works by, well, swapping glyphs and some metadata elsewhere.

1. Should we move instead of swap? I.e. when a rule says `S -> S.closed`, should the content of `S.closed` be moved into `S` and the original `S.closed` be deleted instead of swapping the glyph? I imagine that outline compatibility after swapping isn't guaranteed anymore anyway.

2. What about feature files?

madig:
Not sure I missed something by reducing the component remapping to one loop.

madig:
I think I only need some rounding tests now. Or is there anything else missing?

madig:
Do we need more sanity checking in the `Instantiator` constructor? Do it in fontmake instead? As in:

1. The designspace must be rectangular or whatever the term is
2. Groups in all masters must be the same
3. Glyphs must be compatible (outlines, components, anchors). fontMath does not guarantee to blow up on anchor errors iirc.
4. Anything else?

Lib/fontmake/instantiator.py
```diff
@@ -321,17 +317,15 @@ def collect_glyph_masters(
     Note: skips empty source glyphs if the default glyph is not empty to almost match
     what ufoProcessor is doing. In e.g. Mutator Sans, the 'S.closed' glyph is left
     empty in one source layer. One could treat this as a source error, but ufoProcessor
-    specifically has code to skip that empty glyph and carry on.
+    specifically has code to skip that empty glyph and carry on. The code might be
+    there to support the SourceDescriptor.mutedGlyphNames attribute that clears the
+    glyph from the master that specifies it, we don't support that attribute though.
```
anthrotype: we don't need to be so verbose here, i think you can remove this paragraph.

madig: Maybe for archeology purposes? :)

anthrotype: comments like these will get "rotten" very quickly

madig:
The following code

```Python
def test_interpolation_against_vf(data_dir):
    designspace = designspaceLib.DesignSpaceDocument.fromfile(
        data_dir / "MutatorSans" / "MutatorSans-weight-only.designspace"
    )
    designspace.rules = []
    designspace.loadSourceFonts(ufoLib2.Font.open)
    designspace.instances[0].location = {"weight": 123.456}

    # Generate VF and VF instance.
    vf_designspace = ufo2ft.compileInterpolatableTTFsFromDS(designspace, inplace=False)
    varfont, _, _ = fontTools.varLib.build(vf_designspace)
    varinstance = fontTools.varLib.mutator.instantiateVariableFont(
        varfont, {"wght": 123.456}
    )

    # varfont.save("/tmp/vf.ttf")
    varinstance.save("/tmp/vfinstance.ttf")

    # Generate instance.
    generator = fontmake.instantiator.Instantiator.from_designspace(
        designspace, round_geometry=True
    )
    instance_ufo = generator.generate_instance(designspace.instances[0])
    instance = ufo2ft.compileTTF(instance_ufo, removeOverlaps=False)

    instance.save("/tmp/instance.ttf")
```

Does not write identical TTFs. Some diffs in e.g. `head` and some other misc. tables. The glyf table contains some off-by-off rounding diffs, but mostly diffs seemingly related to cu2qu conversion (more points in curvy glyphs), additional `overlap="1"` point attributes in the VF-cut instance and different component flags. At least the kerning values seem the same. Maybe cutting an instance just isn't the same.

madig:
Todo: add handling of `slnt` axis which translates 1:1 to `post.italicAngle`.

madig:
TestSubset.glyphs or glyphsLib's handling of it is broken. The axis definition comes out as

```
<axes>
    <axis tag="wght" name="Weight" minimum="400" maximum="400" default="400">
      <map input="400" output="700"/>
    </axis>
  </axes>
```

Which trips varLib.

Lib/fontmake/font_project.py
```diff
@@ -36,6 +36,7 @@
 from ufo2ft.featureWriters import FEATURE_WRITERS_KEY, loadFeatureWriters
 from ufo2ft.util import makeOfficialGlyphOrder
 
+import fontmake.instantiator
```
anthrotype: i like `from fontmake import instantiator` better; `fontmake.instantiator` is too long

Lib/fontmake/font_project.py
```diff
@@ -669,50 +670,93 @@ def interpolate_instance_ufos(
...
+            apply_instance_data_to_ufo(instance.font, instance, designspace)
 
-        return instance_ufos
+            if expand_features_to_instances:
+                logger.debug("Expanding features to instance UFO %s", instance.filename)
+                instance.font.features.text = feature_text
+
+            assert (
+                instance.filename is not None
+            ), "FIXME: Make fontmake not rely on instance filenames"
```
anthrotype: what do you mean "not rely on instance filenames"? how does fontmake know where to save the instances if requested to generate them?

madig:
Nice optimization: if the requested location matches a master, just return the master instead of running through varLib. Demonstrated with Noto Sans where the first instance matches a master:

```
In [24]: %timeit g.generate_instance(d.instances[0])                                             
421 ms ± 1.85 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [25]: %timeit g.generate_instance(d.instances[1])                                             
3.63 s ± 64.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

madig:
Note to self: explore Python libraries like Dask and https://pypi.org/project/joblib/ to generate instances in parallel for maximum room heating.

madig:
Interesting performance benefits of Instantiator: Fontmaking NotoSansHebrew-MM.designspace instances takes ~60 seconds on my Ryzen 1700, doing it with Instantiator takes ~20 seconds. Don't know why, but I like it.

madig:
Deepcopying the default source's lib is much slower than iterating over it and assigning key and value to the new dict, but I suppose we have to to make the data independent?

madig:
Designspace loading needs to be refactored. I'd like to sanity-check instance filenames so the code doesn't overwrite sources with instances because the filenames are the same (this has happened to me before) _and_ make sure all instances actually have filenames so I don't have to make one up on the spot when saving instances. Fail fast and early. The `_load_designspace_sources` static method looks like the right place, except it doesn't know the `instance_dir`.

madig:
Idea: with the optimization detailed above, the `-M` switch could be used to apply rules to masters that can't interpolate yet. We currently have a project that is still heavily worked on where interpolation isn't feasible yet, but that has a bracket layer for `dollar`. The team would like to generate the static masters with the `dollar` substitution rule applied so they don't have to resort to `cmap` hacks.

madig:
Idea: pare down core MutatorMath code and bundle it up for use in `instantiator.py`, so users can use either varLib or MutatorMath math, but the UFO assembly code stays the same. This would make it possible to support sparse masters/source layers for MutatorMath.

madig:
Hm. Copying things is tricky. If the Designspace object passed into `Instantiator.from_designspace` has no loaded fonts, we don't need to copy the fonts because we load them ourselves anyway; if there are loaded fonts, we need to `deepcopy` everything everywhere because a cursory glance at fontMath suggests it is not rigorous about maintaining independent copies of the data it handles. Maybe `deepcopy` the passed in Designspace object and then call `loadSourceFonts`?

Lib/fontmake/__main__.py
```diff
@@ -156,6 +156,11 @@ def main(args=None):
         'E.g.: -i "Noto Sans Bold"; or -i ".* UI Condensed". '
         "(for Glyphs or MutatorMath sources only). ",
     )
+    outputGroup.add_argument(
+        "--use-mutatormath",
+        action="store_true",
+        help="Use MutatorMath to generate instances.",
```
anthrotype: maybe mention here in the help section for `--use-mutatormath` why one would like to use that, e.g. if output relies on extrapolation or anisotropic interpolation (not supported by varLib), or why _not_ to use it (doesn't support rules or sparse layers).

Lib/fontmake/font_project.py
```diff
@@ -36,6 +37,7 @@
 from ufo2ft.featureWriters import FEATURE_WRITERS_KEY, loadFeatureWriters
 from ufo2ft.util import makeOfficialGlyphOrder
 
+import fontmake.instantiator as instantiator
```
anthrotype: nit: I prefer `from fontmake import instantiator`

Lib/fontmake/font_project.py
```diff
@@ -646,6 +648,83 @@ def interpolate_instance_ufos(
...
+            )
+            ufo_path_parent = os.path.dirname(ufo_path)
+            if not os.path.exists(ufo_path_parent):
+                os.makedirs(ufo_path_parent)
+            instance.font.save(ufo_path, overwrite=True)
+
+        return [
+            i.font
+            for i in designspace.instances
+            if hasattr(i, "font") and i.font is not None
```
anthrotype: we may want to have InstanceDescriptor.font be initialised to None in upstream designspaceLib (later).

requirements.txt
```diff
@@ -1,8 +1,12 @@
 fonttools[lxml,unicode,ufo]==3.43.0
 cu2qu==1.6.5
-glyphsLib==4.0.0
+#glyphsLib==4.0.0
```
anthrotype: what's missing to use stable glyphsLib version?

Lib/fontmake/font_project.py
madig: What do you mean? `i.font is None` should be the default state until `instance.font = generator.generate_instance(instance)` sets it; this way, we only return UFOs the user selected or all.

anthrotype: exactly. I see only SourceDescriptor has self.font = None in its constructor. InstanceDescriptor should do that too, so you don't need `hasattr(instance, "font")`

requirements.txt
anthrotype: oh right, this is required
https://github.com/googlefonts/glyphsLib/commit/94d68cf3f79597e1c0312cd740f3532b3847b455

as soon as we merge https://github.com/googlefonts/glyphsLib/pull/538, we can cut a new glyphsLib release. I'd like to have that out by end of week preferably.

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,545 @@
...
+            try:
+                glyph_instance = glyph_mutator.instance_at(location_normalized)
+
+                if self.round_geometry:
+                    glyph_instance = glyph_instance.round()
+
+                # onlyGeometry=True does not set name and unicodes, in ufoLib2 we can't
+                # modify a glyph's name. Copy unicodes from default font.
+                glyph_instance.extractGlyph(glyph, onlyGeometry=True)
+            except Exception as e:
```
anthrotype: why such a broad `except Exception` clause? Can't you make it more specific?

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,545 @@
...
+                # By default, explode if we cannot generate a glyph instance for
+                # whatever reason (usually outline incompatibility)...
+                if glyph_name not in self.skip_export_glyphs:
+                    raise ValueError(
```
anthrotype: Maybe it would be good to have our own custom Exception class instead of generic ValueError here

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,545 @@
...
+    def _generate_instance_info(
+        self,
+        instance: designspaceLib.InstanceDescriptor,
+        location_normalized: Location,
+        location: Location,
+        font: ufoLib2.Font,
+    ) -> None:
+        """Generate fontinfo related attributes.
+
+        Seperate, as fontinfo treatment is more extensive than the rest.
```
anthrotype:
```suggestion
        Separate, as fontinfo treatment is more extensive than the rest.
```

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,545 @@
...
+    def from_masters(
+        cls, items: List[Tuple[Location, FontMathObject]], axis_order: List[str]
+    ):
+        masters = []
+        master_locations = []
+        location_to_master = {}
+        for normalized_location, master in items:
+            master_locations.append(normalized_location)
+            masters.append(master)
+            location_to_master[tuple(normalized_location.items())] = master
```
anthrotype: should you maybe sort the `normalized_location.items()` before casting that to a tuple? in case the order happens to be different?

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,545 @@
...
+        model = varLib.models.VariationModel(master_locations, axis_order)
+
+        return cls(masters, location_to_master, model)
+
+    def instance_at(self, normalized_location: Location) -> FontMathObject:
+        # Optimization: If a location matches a master, return the master instead of
+        # running through varLib.
```
anthrotype: this is more than just a simple "optimization". It would allow to have "instances" for masters that are not interpolation-compatible. Think about the way in Glyphs.app you can only ever export instances, never masters as such; so even with a single master font, you always export its implicit "Regular" instance. Instances may carry additional data than the bare masters. With fontmake the `apply_instance_data` step is only done when you pass the `-i` (`--interpolate`) option. However one cannot use that option unless the masters are interpolatable, even if interpolation itself may not be required since all instances defined match the same location as their respetcive masters.
Would be nice to implement this feature in instantiator if it's not too much work. I think it's basically just a matter of identifying the master that correspond to a given instance's location, and returning a copy of its data.

Lib/fontmake/instantiator.py
madig: I don't know what to catch :O

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,545 @@
...
+    copy_lib: Mapping[str, Any]
+    default_design_location: Location
+    designspace_rules: List[designspaceLib.RuleDescriptor]
+    glyph_mutators: Mapping[str, "Variator"]
+    glyph_name_to_unicodes: Dict[str, List[int]]
+    info_mutator: "Variator"
+    kerning_mutator: "Variator"
+    round_geometry: bool
+    skip_export_glyphs: List[str]
+    special_axes: Mapping[str, designspaceLib.AxisDescriptor]
```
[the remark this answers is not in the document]
madig: I found it hard to come up with meaningful defaults, as in, the object can be used in a meaningful way. Maybe wait until a specific use-case arises.

Lib/fontmake/instantiator.py
madig: Mh, yes. Another thing is maybe constructing a location tuple in generate_instance so we don't have to regenerate it every time we call a Variator :thinking:

madig: Urgh, wait, VariationModel expects dict location parameters. Meh. I guess I can attach a `lru_cache` if this turns out to be a memory/performance waster.

Lib/fontmake/instantiator.py
madig: Already possible with the code :+1:

madig: I think this may solve another problem we currently have with a client project: we want to ship a VF and static fonts from the same sources. The VF needs a glyph replacement rule for `dollar` and `cent`, but we currently can't easily apply them to the statics (without also messing with our VTT dumps), so we include both and do a cmap swap hack. This would formalize the hack :)

madig: Mh, this reminds me of my question further up: when replacing glyphs in instances, should we swap or replace the glyphs in question?

anthrotype: oh, I misunderstood the comment, I see that feature is already implemented :)
But I see you're returning the same master object, not a copy. Would that be a problem if following code assumes instanced objects are distinct from the masters'?

> when replacing glyphs in instances, should we swap or replace the glyphs in question?

what's the difference? isn't it the same thing?

madig: (adressed in latest pushes)

Lib/fontmake/__main__.py
```diff
@@ -156,6 +156,14 @@ def main(args=None):
         'E.g.: -i "Noto Sans Bold"; or -i ".* UI Condensed". '
         "(for Glyphs or MutatorMath sources only). ",
     )
+    outputGroup.add_argument(
+        "--use-mutatormath",
```
anthrotype: maybe add a test case in `tests/test_main.py` that uses this `--use-mutatormath` option and verifies it calls mutatormath

Lib/fontmake/font_project.py
anthrotype: how about we `yield` instance UFOs instead of returning a list of them?
the caller may then decide to collect them into a list or to process them one by one

Lib/fontmake/instantiator.py
anthrotype: ok keep it like this for now with a TODO to fix it when we figure out what exceptions fontMath can/should throw

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,579 @@
...
+        if instance.familyName:
+            font.info.familyName = instance.familyName
+        if instance.styleName:
+            font.info.styleName = instance.styleName
+        if instance.postScriptFontName:
+            font.info.postscriptFontName = instance.postScriptFontName
+        if instance.styleMapFamilyName:
+            font.info.styleMapFamilyName = instance.styleMapFamilyName
+        if instance.styleMapStyleName:
+            font.info.styleMapStyleName = instance.styleMapStyleName
```
anthrotype: shouldn't the handling of groups.plist, lib.plist and features.fea be logically separate from the fontinfo.plist proper? here you do all that within the same _generate_intance_info method.

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,579 @@
...
+
+        if not (source_glyph.contours or source_glyph.components):
+            if source is designspace.findDefault():
+                default_glyph_empty = True
+            else:
+                other_glyph_empty = True
+
+        # TODO: What to do about a glyphs lib key? Any useful data that should be
+        # copied from master to all instances? `public.verticalOrigin`, does it
+        # interpolate? `public.postscript.hints`?
```
anthrotype: yes, public.verticalOrigin should be interpolated. Maybe open an issue in fontmath about that

Lib/fontmake/instantiator.py
```diff
@@ -0,0 +1,579 @@
...
+            font.groups[key] = [name for name in glyph_names]
+        font.features.text = self.copy_feature_text
+        # 1. Copy the default lib to the instance.
+        font.lib = typing.cast(dict, copy.deepcopy(self.copy_lib))
+        # 2. Copy the Designspace instance lib to the instance, overwriting the default.
+        for key, value in instance.lib.items():
+            font.lib[key] = copy.deepcopy(value)
+        # 3. Copy the Designspace public.skipExportGlyphs lib key to the instance lib
+        #    to make sure the Designspace version wins.
+        font.lib["public.skipExportGlyphs"] = [name for name in self.skip_export_glyphs]
```
anthrotype: we should let an instance-specific "public.skipExportGlyphs" list from Designspace.instance.lib take precedence over a global one from designspace.lib.

Also this behavior whereby designspace.instances.lib elements are copied to the generated UFO instance should be documented in the Designspace spec (I don't think it is, currently, beyond a generic "custom data for this instance").

Lib/fontmake/font_project.py
madig: The MM code path returns a list, maybe for a later refactor PR?

Lib/fontmake/instantiator.py
madig: Done: https://github.com/robotools/fontMath/issues/161

Still drop glyph lib key for now?

Lib/fontmake/instantiator.py
madig: 1. Done
2. https://github.com/fonttools/fonttools/pull/1683

Lib/fontmake/instantiator.py
madig: Yes.

Lib/fontmake/font_project.py
madig: My concern is what happens if `filename` is not specified.

anthrotype:
nice work @madig 👏

Issue #560, opened 2019-06-17
next fontmake version will require python 3.6 or greater


anthrotype:
it's about time.
@madig is working on that, thanks Nikolaus

anthrotype:
Fixed by #561

PR #561, opened 2019-06-17
Make Python 3 only

madig:
https://github.com/googlefonts/fontmake/issues/560

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -12,9 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-from __future__ import absolute_import, division, print_function, unicode_literals
-
 import glob
 import logging
 import math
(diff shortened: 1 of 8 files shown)


Lib/fontmake/font_project.py
```diff
@@ -1179,7 +1162,7 @@ def buildTables(self):
             stdout, stderr = process.communicate()
             retcode = process.poll()
 
-            report = tostr(stdout + (b"\n" + stderr if stderr else b""))
+            report = str(stdout + (b"\n" + stderr if stderr else b""))
```
anthrotype: don't call `str()`, but do the decoding to 'ascii', since the output from subprocess is always bytes.

```python
(stdout + (b"\n" + stderr if stderr else b"")).decode("ascii")
```

anthrotype: if you were to do `str(b"a")` you'd get the bytes repr "b'a'" back, which is not what you want.

anthrotype: or if you want to keep calling `str()` you need to pass the `encoding="ascii"` keyword argument so that it in turn calls the decode method for you.

Issue #567, opened 2019-08-06
Instantiator: revisit deepcopying of data


madig:
On Noto Sans, generating the first two instances initially took

```
In [24]: %timeit g.generate_instance(d.instances[0])                                             
421 ms ± 1.85 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [25]: %timeit g.generate_instance(d.instances[1])                                             
3.63 s ± 64.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Now it takes

```
In [8]: %timeit i.generate_instance(d.instances[0])
985 ms ± 6.11 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: %timeit i.generate_instance(d.instances[1])
3.2 s ± 11.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

1. Audit fontMath's `extract*` and `round()` methods -- can we drop deepcopying in `Variator.instance_at`?
2. Maybe add `inplace` parameter when we know that having data interlinked (when reading and writing to disk) is harmless?

madig:
I've been looking at the `extract*` and `round()` methods and followed the code, and they seem to be making copies where necessary (primarily lists, dicts and complex objects). So I'd say we don't need to deepcopy masters if an instance location matches a master.

madig:
I suppose this has to wait until we can ensure that fontMath or a replacement always returns copies and never modifies internal data.

PR #575, opened 2019-08-28
Instantiator: don't copy actual masters

madig:
Close https://github.com/googlefonts/fontmake/issues/567.

We only call `.round()` and `extract*()` on instantiated fontMath objects, and fontMath seems to make copies appropriately, so we should not need to `deepcopy` everything.


madig:
MathKerning actually works on the instance's data directly. Meh. The copy must stay then...

Closing this until we can ensure that all the methods we call return copies.

behdad:
Why haven't we replaced fontMath yet?

madig:
Available man power? :)

behdad:
We should have been more proactive recruiting women. :)

PR #578, opened 2019-09-09
Copy font info attributes from default font to instance selectively

### LINKED ISSUE #574: Instance generation: warn when default font contains font info entries that are instance-specific and are not overridden in the Designspace instance attributes
I recently debugged a release where all instances were turned into "Regular". It turned out that the default font contained `openTypeNamePreferred*` font info fields that made ufo2ft turn everything into the default font name-wise even though instance style names etc. were present.

The following fields should probably not be in the `default_font.info` when generating instances, unless they're being overridden by Designspace `InstanceDescriptor` attributes. This can confuse ufo2ft or just generate the wrong data in the instances. Print a warning.

- postscriptFontName
- styleMapFamilyName
- styleMapStyleName
- styleName

The following attributes can't be set directly and would have to be set in e.g. a `com.schriftgestaltung.customParameters` key:

- macintoshFONDName
- openTypeNameCompatibleFullName
- openTypeNamePreferred
- openTypeNamePreferredSubfamilyName
- openTypeNameUniqueID
- openTypeNameWWSFamilyName
- openTypeNameWWSSubfamilyName
- openTypeOS2Panose
- postscriptFullName
- postscriptUniqueID
- postscriptWeightName
- woffMetadataUniqueID
- (italicAngle? #577)

What about openTypeNameRecords in the default font? Sniff out the nameIDs of styleMapFamilyName, styleMapStyleName? What about styleName?

madig:
Closes #574.

I overlooked that ufoProcessor does the same. Several attributes are actually specific to instances and should not be blindly inherited from the default font.

The ignore list is different from ufoProcessor. We additionally ignore:
- openTypeNameUniqueID
- openTypeNameWWSFamilyName
- openTypeNameWWSSubfamilyName
- openTypeOS2Panose

We copy over:
- familyName
- woffMajorVersion
- woffMetadataCopyright
- woffMetadataCredits
- woffMetadataDescription
- woffMetadataExtensions
- woffMetadataLicense
- woffMetadataLicensee
- woffMetadataTrademark
- woffMetadataVendor
- woffMinorVersion

This implies that the styleName attribute is now implicitly required. Not specifying it in the Designspace means it will not be set in the instance.

### DIFF
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -102,6 +102,9 @@ WDTH_VALUE_TO_OS2_WIDTH_CLASS = {
 # - macintoshFONDFamilyID
 # - macintoshFONDName
 # - year
+#
+# This means we implicitly require the `stylename` attribute in the Designspace
+# `<instance>` element.
 UFO_INFO_ATTRIBUTES_TO_COPY_TO_INSTANCES = {
     "copyright",
     "familyName",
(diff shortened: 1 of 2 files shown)


madig:
Hm. I'll add a warning if `stylename` isn't set.

PR #599, opened 2019-11-18
use ufoLib2 to load UFOs

anthrotype:
Should be a bit faster. E.g. generating the master UFOs from NotoSans-MM.glyphs goes from 39 seconds to 18 on my local machine.

Also make defcon and mutatormath not required by default, guarded behind an extra installation flag (e.g. `pip install fontmake[mutatormath]`.


Lib/fontmake/font_project.py
```diff
@@ -109,8 +108,13 @@ def __init__(self, timing=False, verbose="INFO", validate_ufo=False):
         logger.debug(
             "ufoLib UFO validation is %s", "enabled" if validate_ufo else "disabled"
         )
-        setUfoLibReadValidate(validate_ufo)
-        setUfoLibWriteValidate(validate_ufo)
+        self.validate_ufo = validate_ufo
```
madig: 👍

PR #632, opened 2020-01-29
instantiator: when applying DS rules, must also swap anchors

anthrotype:
Glyph.clear() destroys all anchors (and pen protocol does not 'draw' them).
`fontmake.instantiator` needs to also swap the glyph anchors when applying the DesignSpace conditional substitution rules, just like it does for the contours and components.
Otherwise, the swapped glyphs will lose all their anchors and get no mark/mkmk features...

**NOTE**: If you used fontmake to generate fonts that contain "bracket" glyphs or a DesignSpace document with rvrn-style conditional substitution rules, and these glyphs contain any anchors which were supposed to be used to generate mark or mkmk feature, I strongly recommend that you re-generate your fonts using the new fontmake and glyphsLib (which we are going to release shortly after this is merged).

Apologies, we should have caught this earlier.
Many thanks @chrissimpkins and @danielgrumer for reporting this bug (and this one too https://github.com/googlefonts/glyphsLib/issues/578).

### DIFF
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -582,17 +582,24 @@ def swap_glyph_names(font: ufoLib2.Font, name_old: str, name_new: str):
     glyph_old.drawPoints(p)
     glyph_swap.width = glyph_old.width
 
-    glyph_old.clear()
+    glyph_old.clearContours()
+    glyph_old.clearComponents()
     p = glyph_old.getPointPen()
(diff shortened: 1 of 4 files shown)


marekjez86:
thanks @anthrotype @madig

danielgrumer:
Thanks guys! 👏👏

chrissimpkins:
Thanks @anthrotype & @madig!!

PR #643, opened 2020-03-27
Add --filter-instances argument

m4rc1e:
Users can now specify which instances they would like to keep when generating a variable font. It works in exactly the same manner as generating static fonts.

@anthrotype there's no rush for this. If you're happy with the implementation,  I can add a test and update the doc strings if needed.

I've also fixed the `_search_instances` func.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -144,7 +144,8 @@ def main(args=None):
         help="Output folder. By default, output folders are created in the "
         "current working directory, grouping output fonts by format.",
     )
-    outputGroup.add_argument(
+    outputInstanceGroup = outputGroup.add_mutually_exclusive_group()
+    outputInstanceGroup.add_argument(
         "-i",
         "--interpolate",
         nargs="?",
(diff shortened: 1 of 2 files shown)


anthrotype:
I don't know if it's a good idea, even though I see how it may be convenient in some situations..
We already sort of abused the existing `-i` option (which originally was meant to select the mode fontmake is running on: ie. build only masters [which is debatable whether it makes sense for binary outputs], or build instances) when we decided to overload it with the ability to filter out the instances to be generated. It kind of made sense for the static outputs.
But for the "intrepolatable" outputs, I think this would only create confusion.
What is the meaning of passing, e.g., `-i` without arguments for `variable` or `interpolatable-ttf` outputs. No-op?
Maybe, if we really want to be able to filter the named instances from the CLI, we should think about a new option, distinct from `-i/--interpolate`.

m4rc1e:
Agreed.

I'm happy to add this functionality as an additional argument to the CLI? I'm also more than happy to explain further if you think this feature is a bad idea in general.

anthrotype:
it's not that it's a bad idea per se, my concern is that the reuse of the existing option may create confusions on the meaning of that option.

m4rc1e:
How about a cli arg like `--filter-instances`?

anthrotype:
ok, let's add `--filter-instances` (which takes a required argument, unlike `-i` where arg is optional).
And let's make that new option work for both static and intrepolatable/variable outputs.
The `-i` option remains excluded for variable outputs.
In the unlikely event that one passes both `-i` with an argument _and_ this new `--filter-instances`, then raise an argument error.

m4rc1e:
Thanks. Let's keep this pr open, I'll make the update in the next few days.

m4rc1e:
Still wip. I'll let you know once I'm happy with it.

m4rc1e:
I'm going to close this since I'm likely to be the only user who wants this functionality.

My current workaround involves converting a .glyphs family to ufos and then generating the family using two .designspaces. One for statics, the other for the VF.

davelab6:
> using two .designspaces

This seems like a failure of the `.designspace` schema; a class attribute for elements would allow a single design space, that would be filtered by such an argument. 

This will be needed for all GF projects as we want to be able to generate VFs with a limited number of fvar named instances, a full set of STAT table instances, and an in-between set of static fonts.

PR #664, opened 2020-05-12
Ensure only the default groups are used in interpolation

madig:
The intention (and varLib's behavior probably) is to only use the default source's groups for kerning group information. Before this PR, we accidentally had Instantiator use the groups associated with each source individually. fontMath makes a union of them, which is fine if all groups are the same, but leads to problems otherwise.

ufoProcessor suffers from the same problems and assumes groups are identical: https://github.com/LettError/ufoProcessor/blob/88323a4/Lib/ufoProcessor/__init__.py#L392-L393

### DIFF
--- a/tests/test_instantiator.py
+++ b/tests/test_instantiator.py
@@ -60,6 +60,33 @@ def test_interpolation_weight_width_class(data_dir):
     assert font.info.openTypeOS2WidthClass == 9
 
 
+def test_default_groups_only(data_dir):
+    d = designspaceLib.DesignSpaceDocument()
+    d.addAxisDescriptor(
+        name="Weight", tag="wght", minimum=300, default=300, maximum=900
+    )
+    d.addSourceDescriptor(location={"Weight": 300}, font=ufoLib2.Font())
+    d.addSourceDescriptor(location={"Weight": 900}, font=ufoLib2.Font())
+    d.addInstanceDescriptor(styleName="2", location={"Weight": 400})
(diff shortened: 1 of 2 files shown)


tests/test_instantiator.py
```diff
@@ -60,6 +60,63 @@ def test_interpolation_weight_width_class(data_dir):
...
+    """Test that only the default source's groups end up in instances."""
+
+    d = designspaceLib.DesignSpaceDocument()
+    d.addAxisDescriptor(
+        name="Weight", tag="wght", minimum=300, default=300, maximum=900
+    )
+    d.addSourceDescriptor(location={"Weight": 300}, font=ufoLib2.Font())
+    d.addSourceDescriptor(location={"Weight": 900}, font=ufoLib2.Font())
+    d.addInstanceDescriptor(styleName="2", location={"Weight": 400})
+    d.findDefault()
```
anthrotype, line 73: Is this necessary?

madig: Yes, so set the `default` property.

madig: Mh. It is done on load in dsLib usually, just not when we construct the objects ourselves.

anthrotype: We should make that default a cached property

madig: It needs to be recomputed every time you change sources or axis data. Maybe making it a simple just-in-time lookup is enough?

PR #668, opened 2020-05-21
Use default glyphs for interpolation

madig:
The default font determines which glyphs are interpolated, because the math behind varLib and MutatorMath needs the default font as a point of reference for all data to be interpolated.

### DIFF
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -192,14 +192,14 @@ class Instantiator:
 
         designspace.loadSourceFonts(ufoLib2.Font.open)
 
-        # The default font determines which glyphs are interpolated, because the math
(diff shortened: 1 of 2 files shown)


Lib/fontmake/instantiator.py
```diff
@@ -192,9 +192,25 @@ def from_designspace(
...
+        default_font = designspace.default.font
+        glyph_names: Set[str] = set(default_font.keys())
+
         for source in designspace.sources:
-            glyph_names.update(source.font.keys())
+            other_names = set(source.font.keys())
+            diff_names = other_names - glyph_names
+            if diff_names:
+                logger.warning(
+                    "The source %s (%s) has more glyphs than the default source, which "
```
anthrotype, line 206:
```suggestion
                    "The source %s (%s) contains glyphs that are missing from the default source, which "
```

PR #671, opened 2020-06-05
support subroutinizing with cffsubr; add otf-cff2 static output

anthrotype:
Uses the latest ufo2ft with cffsubr support to allow subroutinizing CFF2 tables.
Also adds a new 'otf-cff2' output format, for non-variable CFF2 fonts

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -1050,8 +1050,8 @@ class FontProject:
             )
 
         need_reload = False
-        if "otf" in output or "otf-cff2" in output:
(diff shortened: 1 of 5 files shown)


Lib/fontmake/font_project.py
```diff
@@ -1041,8 +1050,9 @@ def run_from_ufos(self, ufos, output=(), **kwargs):
             )
 
         need_reload = False
-        if "otf" in output:
-            self.build_otfs(ufos, **kwargs)
+        if "otf" in output or "otf-cff2" in output:
```
madig, line 1053: maybe unroll this one, the double-if is confusing to read

PR #689, opened 2020-08-20
Fix two problems: non-determinism and incorrect rule application in instanciator

belluzj:
Hello, I pushed here a test that reproduces a bug with the instanciator when several designspace rules touch the same glyphs. I wrote a test designspace and [a description of what should happen in the docstring of the test](https://github.com/googlefonts/fontmake/pull/689/files#diff-60bba4bf61f2028edf91b4e75a5728f2R276-R288).

Instead of the expected result, there are two problems I would like to address in this PR:
* with the rules of the designspace provided as a test case, the output is wrong
* the output is also random, which can make it right sometimes by luck

The misbehaved code is below:
https://github.com/googlefonts/fontmake/blob/f07b1932768efdee8f15a4a9179d1c2d181ceb8a/Lib/fontmake/instantiator.py#L376-L383

Here is what I get if I put a breakpoint after line 380:

```
$ fontmake -m MyFont.designspace -o ttf -i
INFO:fontmake.font_project:Interpolating master UFOs from designspace
INFO:fontmake.font_project:Generating instance UFO for "My Font Style Black"
> c:\userslocal\jany.belluz\documents\code\fontmake\lib\fontmake\instantiator.py(382)generate_instance()
-> for name_old, name_new in zip(glyph_names_list, glyph_names_list_renamed):
(Pdb) glyph_names_list
dict_keys(['Q.ss01.alt', 'Q.ss01', 'Q.alt', 'Q'])
(Pdb) glyph_names_list_renamed
['Q.ss01.alt', 'Q.ss01.alt', 'Q.ss01.alt', 'Q.ss01.alt']
```

As you can see from the contents of `glyph_names_list` and `glyph_names_list_renamed`, the `for` loop will perform the following swaps:
1. Q.ss01 <-> Q.ss01.alt
2. Q.alt <-> Q.ss01.alt
3. Q <-> Q.ss01.alt

#### Initial
| name | outlines |
|---|---|
| Q | Q |
| Q.alt | Q.alt |
| Q.ss01 | Q.ss01 |
| Q.ss01.alt | Q.ss01.alt |

#### Step 1
| name | outlines |
|---|---|
| Q | Q |
| Q.alt | Q.alt |
| Q.ss01 | Q.ss01.alt |
| Q.ss01.alt | Q.ss01 |

#### Step 2
| name | outlines |
|---|---|
| Q | Q |
| Q.alt | Q.ss01 |
| Q.ss01 | Q.ss01.alt |
| Q.ss01.alt | Q.alt |

#### Step 3 = final
| name | outlines |
|---|---|
| Q | Q.alt |
| Q.alt | Q.ss01 |
| Q.ss01 | Q.ss01.alt |
| Q.ss01.alt | Q |

The end result is not right, because what we would want is that the Q gets the outlines of Q.ss01.alt.

And if you run the command again you get a different result, because `glyph_name_list` ~~is keys from a dict~~ comes from a `set()` at some point, so the order is random, see example below:

```
$ fontmake -m MyFont.designspace -o ttf -i
INFO:fontmake.font_project:Interpolating master UFOs from designspace
INFO:fontmake.font_project:Generating instance UFO for "My Font Style Black"
> c:\userslocal\jany.belluz\documents\code\fontmake\lib\fontmake\instantiator.py(382)generate_instance()
-> for name_old, name_new in zip(glyph_names_list, glyph_names_list_renamed):
(Pdb)  glyph_names_list
dict_keys(['Q.ss01.alt', 'Q', 'Q.alt', 'Q.ss01'])        # <--- different order here
```

Thinking about this by hand, here is what I think would make sense as a series of swaps (which is basically applying the designspace rules one after the other in order):

#### Initial
| name | outlines |
|---|---|
| Q | Q |
| Q.alt | Q.alt |
| Q.ss01 | Q.ss01 |
| Q.ss01.alt | Q.ss01.alt |

#### Apply [rule 1](https://github.com/googlefonts/fontmake/pull/689/files#diff-896394abc6db517c2655cab987b3b6c4R13): Q <-> Q.alt
| name | outlines |
|---|---|
| Q | Q.alt |
| Q.alt | Q |
| Q.ss01 | Q.ss01 |
| Q.ss01.alt | Q.ss01.alt |

#### Apply [rule 2](https://github.com/googlefonts/fontmake/pull/689/files#diff-896394abc6db517c2655cab987b3b6c4R20): Q.ss01 <-> Q.ss01.alt
| name | outlines |
|---|---|
| Q | Q.alt |
| Q.alt | Q |
| Q.ss01 | Q.ss01.alt |
| Q.ss01.alt | Q.ss01 |

#### Apply [rule 3](https://github.com/googlefonts/fontmake/pull/689/files#diff-896394abc6db517c2655cab987b3b6c4R27-R28): Q <-> Q.ss01 and Q.alt <-> Q.ss01.alt
| name | outlines |
|---|---|
| Q | Q.ss01.alt |
| Q.alt | Q.ss01 |
| Q.ss01 | Q.alt |
| Q.ss01.alt | Q |

This gives a satisfying result for the main case (Q has the outlines of Q.ss01.alt) and also for the others, I feel, since the final swaps make sense (the style is changed so the ss01 has been swapped, and the weight is bold enough so the alt have been swapped). Do you agree that I should implement this result?

Next question: if I want to implement the result above, I need more information than what `designspaceLib.processRules()` returns, so I would need either to modify what that function returns (is that alright? Do other libraries rely on the previous output? I guess yes) or introduce another functions that returns what I want.

@anthrotype what do you think? Sorry for the long explanation!

### DIFF
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -153,7 +153,7 @@ class InstantiatorError(Exception):
     pass
 
 
-def processRulesSwaps(rules, location, glyphNames):
+def process_rules_swaps(rules, location, glyphNames):
     """ Apply these rules at this location to these glyphnames
         - rule order matters
 
@@ -397,7 +397,7 @@ class Instantiator:
         glyph_names_list = self.glyph_mutators.keys()
         # The order of the swaps below is independent of the order of glyph names.
         # It depends on the order of the <sub>s in the designspace rules.
-        swaps = processRulesSwaps(self.designspace_rules, location, glyph_names_list)
+        swaps = process_rules_swaps(self.designspace_rules, location, glyph_names_list)
         for name_old, name_new in swaps:
             if name_old != name_new:
                 swap_glyph_names(font, name_old, name_new)
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -153,6 +153,26 @@ class InstantiatorError(Exception):
     pass
 
 
+def processRulesSwaps(rules, location, glyphNames):
+    """ Apply these rules at this location to these glyphnames
+        - rule order matters
+
+        Return a list of (oldName, newName) in the same order as the rules.
+    """
+    swaps = []
+    glyphNames = set(glyphNames)
+    for rule in rules:
+        if designspaceLib.evaluateRule(rule, location):
+            for oldName, newName in rule.subs:
+                # Here I don't check if the new name is also in glyphNames...
+                # I guess it should be, so that we can swap, and if it isn't,
+                # then it's better to error out later when we try to swap,
+                # instead of silently ignoring the rule here.
+                if oldName in glyphNames:
+                    swaps.append((oldName, newName))
+    return swaps
+
+
 @attr.s(auto_attribs=True, frozen=True, slots=True)
 class Instantiator:
     """Data class that holds all necessary information to generate a static
@@ -375,10 +395,10 @@ class Instantiator:
 
         # Process rules
         glyph_names_list = self.glyph_mutators.keys()
-        glyph_names_list_renamed = designspaceLib.processRules(
-            self.designspace_rules, location, glyph_names_list
-        )
-        for name_old, name_new in zip(glyph_names_list, glyph_names_list_renamed):
+        # The order of the swaps below is independent of the order of glyph names.
+        # It depends on the order of the <sub>s in the designspace rules.
+        swaps = processRulesSwaps(self.designspace_rules, location, glyph_names_list)
+        for name_old, name_new in swaps:
             if name_old != name_new:
                 swap_glyph_names(font, name_old, name_new)
 
--- a/tests/data/DesignspaceRuleOrder/MyFont.designspace
+++ b/tests/data/DesignspaceRuleOrder/MyFont.designspace
@@ -1,0 +1,61 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<designspace format="4.1">
+    <axes>
+        <axis tag="wght" name="Weight" minimum="100" maximum="900" default="100" />
+        <axis tag="STYL" name="Style" minimum="0" maximum="1" default="0"/>
+    </axes>
+
+    <rules processing="last">
+        <rule>
+            <conditionset>
+                <condition name="Weight" minimum="780"/>
+            </conditionset>
+            <sub name="Q" with="Q.alt"/>
+        </rule>
+
+        <rule>
+            <conditionset>
+                <condition name="Weight" minimum="730"/>
+            </conditionset>
+            <sub name="Q.ss01" with="Q.ss01.alt"/>
+        </rule>
+
+        <rule>
+            <conditionset>
+                <condition name="Style" minimum="0.5" maximum="1"/>
+            </conditionset>
+            <sub name="Q" with="Q.ss01"/>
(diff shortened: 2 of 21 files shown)


belluzj:
@anthrotype I mis-clicked to publish the PR before I had time to write the description, I wanted to make sure you got a notification now that I'm done writing.

@letterror I think you wrote `designspaceLib.processRules()`, what do you think about my proposed handling of the case above? Could that function be changed to return, for example, an ordered list of pairs of glyphs to swap? (or anything that allows me to implement the case above properly)

Thanks in advance for your advice.

anthrotype:
Thanks Jany! I may not have time to look into this until the second week of September when I'll be back from holidays.

From a very quick check, I find a bit surprising that the dict keys are non deterministic, my understanding was that in Python >= 3.6 dictionaries keep items in insertion order. /cc @madig who authored this particular piece of code in fontmake

anthrotype:
hm, this for loop that populates self.glyph_mutators iterates over a set (whose order is random even in python 3):

https://github.com/googlefonts/fontmake/blob/f07b1932768efdee8f15a4a9179d1c2d181ceb8a/Lib/fontmake/instantiator.py#L248

The glyph_names set is created here
https://github.com/googlefonts/fontmake/blob/f07b1932768efdee8f15a4a9179d1c2d181ceb8a/Lib/fontmake/instantiator.py#L199

anthrotype:
processRules's docstrings emphasizes that "rules order matters", but I don't see why the order of the glyphNames should

https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/designspaceLib/__init__.py#L205-L224

belluzj:
Yes, I think the random ordering is purely a fontmake problem and has nothing to do with processRules()... I think I will fix the randomness here and also implement an alternative version of processRules here so that we can see how it could be different and comment on it, and later we can discuss if the alternative implementation should be upstreamed to designspaceLib somehow.

Lib/fontmake/instantiator.py
```diff
@@ -153,6 +153,26 @@ class InstantiatorError(Exception):
     pass
 
 
+def processRulesSwaps(rules, location, glyphNames):
```
anthrotype, line 156: minor: the rest of the function names in fontmake.instantiator use snake_case, might be worth doing the same for process_rules_swaps

anthrotype (approved):
this looks good to me

anthrotype:
Ideally we'd also have your processRulesSwaps in the upstream designspaceLib -- but I don't think that should block this.

I think ufoProcessor suffers from the same issue because it uses the current processRules where the order of glyphNames matters, however this glyphNames is created by turning a set into a list (hence order is random). /cc @LettError

LettError:
What do you propose?

anthrotype:
> What do you propose?

the proposed `processRulesSwaps` in this PR solves the issue by returning a list of (oldName, newName) swaps ordered by rule older that doesn't depend on the ordering of the `glyphNames` list (which originally being a set has a random order). We should add a similar function to designspaceLib, or modify the current designspaceLib.processRules to do the same.

belluzj:
Should I merge this?

belluzj:
Oh actually I can't merge on this repository sorry! Please do if you're happy.

anthrotype:
> I can't merge on this repository

I just sent you an invite, please try again then -- thanks

PR #707, opened 2020-11-10
Switch to GitHub Actions

madig:


.github/workflows/test.yml
```diff
@@ -0,0 +1,25 @@
...
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.platform }}
+    strategy:
+      matrix:
+        python-version: [3.6, 3.9]
```
alerque, line 14: Is there a reason to drop testing for 3.7 and 3.8? I don't know about these Python versions, but for some other ecosystems it can be a lifesaver to get results over a full matrix of supported version ranges.

alerque (approved):

.github/workflows/test.yml
madig: Simplicity/laziness mainly. It it works on the oldest and newest supported version, it probably works in-between.

madig:
GitHub Actions' Windows runners have a funny tendency to expose platform bugs.

.github/workflows/test.yml
alerque: My experience with Python is than in-between versions sometimes have regressions that need extra work arounds. It's been a while since I've used it actively enough to track things like that, but given [the huge breadth](https://repology.org/project/python/versions) of versions in distros out there I'd be inclined to go ahead and err on the side of testing a few more in-betweens.

anthrotype: I think it's worth testing them all from 3.6 through to 3.9

Issue #660, opened 2020-05-03
Option for external feature file?


simoncozens:
I'm trying to build a font which a Glyphs source for the outlines/metrics/info but an external .fea file for layout. I could obviously do it in two steps (fontmake to create OTF, then feaLib to add features) but I'm surprised there isn't an obvious way to do this from fontmake. Confusingly the options say:

```
  --mti-source MTI_SOURCE
                        Path to mtiLib .txt feature definitions (use instead
                        of FEA)
```

But there is no other reference to FEA.

madig:
There's no obvious way because both UFO and Glyphs formats package feature files inside, so this is an unsupported edge case. The MTI switch is just bolted on because some Noto fonts have MTI source feature files.

anthrotype:
you could try using an `include` statement that references the external feature file. That's supposed to work

twardoch:
Include works in many cases :)

simoncozens:
OK, thanks - got it working with a reference in the Glyphs file and `--master-dir .` to make references resolve.

PR #709, opened 2020-11-12
add "--no-generate-GDEF" command line parameter

### LINKED ISSUE #702: ValueError: The features already contain a `table GDEF {...}` statement
Hi there,

fontmake currently fails to build the Angular variant of the comic-sans font from the glyphs-format into OTF files.

The font can be found here:
https://github.com/crozynski/comicneue

The command I use to compile the font is:
```
fontmake -o otf -g Sources/Glyphs/Files-from-UFO/Angular-from-UFO/ComicNeue-Angular-Regular.glyphs 
```

However, it fails with the following error message:
```
ValueError: The features already contain a `table GDEF {...}` statement. Either delete it or set generate_GDEF to False.
```

I mean, if the GDEF table is already there, fine, why not keep it? So, how do I set `generate_GDEF` to False or, alternatively, how do I tell fontmake to delete and re-generate the table and why isn't this done by default?

This is on Debian testing/unstable with the following packages installed:

fontmake 2.1.4-1
python3-cu2qu 1.6.7-1+b2
python3-fontmath 0.6.0-1
python3-fonttools 4.16.1-1
python3-glyphslib 5.2.0+ds1-1
python3-ufo2ft 2.14.1-1
python3-ufolib2 0.8.0+dfsg1-1
ttfautohint 1.8.3-2+b1

fabiangreffrath:
This adds the "--no-generate-GDEF" command line parameter which sets
the "generate_GDEF" variable which in turn is passed over to
glyhpsLib invocations to determine whether to auto-generate a GDEF
table or not.

Fixes #702.


fabiangreffrath:
@googlebot I signed it!

anthrotype:
Thanks. A test would be nice too.
Can you please remind me what is the intended use case?
IIUC, you have a .glyphs input source where the features text contains some human-written `table GDEF {...};` code, and you don't want glyphsLib to generate a new GDEF. Is that correct?

What kind of data does your GDEF contain?

What does Glyphs.app do in this case? Does it silently discard the user-written GDEF table and regenerate a new one? Does it keep the user GDEF verbatim? Or a mix of the two? Does it have any custom parameter to control this?

fabiangreffrath:
@anthrotype My use case is mostly what is described in #702 : I want to compile ComicNeueAngular from its glyphs source, but this unexpectedly contains a GDEF table and thus fontmake bombs out with an error inherited from glyphsLib. I found no way to tell either fontmake or glyphsLib "if you find an unexpected GDEF table, just discard it a generate a new one". But what I found is the "generate_GDEF" parameter to `glyphsLib.to_designspace` so I use this one to tell glyphsLib through fontmake not to generate a new GDEF table at all.

anthrotype:
right, but "discarding and generating a new one" is not the same as "not to generate a new GDEF table at all"..

anthrotype:
> unexpectedly contains a GDEF table

the unexpectedly is the problem here. Was it put there intentionally? If so, why? Is it a leftover from a previous conversion? If so, i'd recommend to re-generate the source file that contains the unexpected GDEF table so that it does not contain one (and it gets regerenated).

The reason glyphsLib doesn't automatically regenerate one is to prevent unexpected data losses.

fabiangreffrath:
I see. So what we need is an option to force glyphsLib to override and re-generate an existing GDEF table instead of erroring out.

moyogo:
> I see. So what we need is an option to force glyphsLib to override and re-generate an existing GDEF table instead of erroring out.

Wouldn't the better approach be to modify the font sources?

simoncozens:
> Wouldn't the better approach be to modify the font sources?

Not necessarily, no. Quoting from a discussion in https://github.com/simoncozens/fontFeatures/issues/35 : 

> Reflecting some more on this: the fontmake answer would be "fontmake is getting GDEF wrong because your Glyphs sources are wrong". And there's some merit in that. U+1BE7 looks very much like a mark but we haven't said that we want it to be a base, even though (looking at the Noto binary) we do want it to be a mark.
>
> But the fontmake answer would continue "You need to fix your sources. If you want U+1BE7 to be a base instead of a mark, then you need to set it as a base in the Glyphs source by going to the Glyph Info window and setting Category to Letter, otherwise you will get the auto-guessed default."
>
> Which is certainly one answer.
>
> But my answer would be, "No, I am the font engineer, and sometimes designers send over bad sources, and I need to deal with that situation. And the mechanism I have for dealing with that is by putting my own GDEF definitions in the feature file, which is the bit of the process under my control. So when I do put my own definitions in the feature file, fontmake should actually use them."
>
> In other words, I think the idea in googlefonts/fontmake#709 for adding a flag to stop fontmake generating its own GDEF tables is a good one, and will solve your problem if it gets merged.

I'm in support of this PR.

anthrotype:
Ok then. Can your rebase and fix the merge conflicts?

fabiangreffrath:
Done.

anthrotype:
@moyogo made me think that we don't actually need to expose a command line option to disable GDEF generation in glyphsLib. Fontmake could simply check if a GDEF table is already present, just before calling glyphsLib to build the master UFOs, and set the generate_GDEF accordingly: i.e. False if one is already present, True if not.

simoncozens:
In #660 I asked about adding a separate feature file, and the advice I was given then was to put an include statement in the Glyphs feature prefix to include an external file.

So by all means check if there already is a GDEF table defined in the feature code - but make sure you do it properly...

anthrotype:
yes. We need to do something similar to `ufo2ft.featureCompiler.parseLayoutFeatures` (using feaLib Parser), but instead of on UFOs, on a .glyphs file 

https://github.com/googlefonts/ufo2ft/blob/master/Lib/ufo2ft/featureCompiler.py#L24

anthrotype:
well, adding an option like the one proposed here would definitely be easier than having to follow includes and parse the feature file just to be able to say whether a GDEF was already defined, and the also throw away the result of that parsing... I'm inclined to go on and accept this PR as is, at least until we figure out a way to generate GDEF table from UFO new glyph properties in ufo2ft. /cc @moyogo @madig @belluzj

moyogo:
Having another feature file included with an include statement in the Glyphs file is also an issue that needs to be addressed in glyphsLib when using `glyphs2ufo --generate-gdef`.

belluzj:
I'm ok with it, as it's just a new option, it doesn't change default behaviour, right? The better smarter version that does the right thing can come later.

anthrotype:
it doesn't change the default behaviour, no. Ok, i'll merge this now. Thanks @fabiangreffrath

PR #714, opened 2020-12-07
Add flatten components option

simoncozens:
Requires googlefonts/ufo2ft#438

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -259,6 +259,13 @@ def main(args=None):
         "measured in EM",
     )
     contourGroup.add_argument(
+        "-f",
+        "--flatten-components",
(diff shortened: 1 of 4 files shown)


anthrotype:
ufo2ft v2.19.1 with `flattenComponents` has been released you may want to require `ufo2ft>=2.19.1` in setup.py's `install_requires`, and also fix to the exact version in requirements.txt (`ufo2ft==2.19.1`)

anthrotype:
Also you should rebase off origin/master, because I have made some changes to the CI setup

anthrotype:
@simoncozens you probably also need to bump minimum fonttools as the latest ufo2ft may require a more recent version than the one currently pinned in fontmake's requirements

anthrotype (approved):

PR #737, opened 2021-02-18
Add --filter argument

moyogo:
This adds a `--filter` argument.
Depends on https://github.com/googlefonts/ufo2ft/pull/466 and https://github.com/googlefonts/ufo2ft/pull/479.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -26,34 +26,33 @@ from fontmake.errors import FontmakeError
 from fontmake.font_project import INTERPOLATABLE_OUTPUTS, FontProject
 
 
-def _loadFeatureWriters(parser, specs):
-    feature_writers = []
+def _loadPlugins(parser, specs, from_string_func, parser_error_message):
+    plugins = []
     for s in specs:
         if s == "None":
-            # magic value that means "don't generate any features!"
+            # magic value that means "don't write any features or don't apply
+            # any filters!"
             return []
         try:
-            feature_writers.append(loadFeatureWriterFromString(s))
+            plugins.append(from_string_func(s))
         except Exception as e:
-            parser.error(
-                "Failed to load --feature-writer:\n  {}: {}".format(type(e).__name__, e)
-            )
-    return feature_writers
+            parser.error(parser_error_message.format(type(e).__name__, e))
+    return plugins
+
+
+def _loadFeatureWriters(parser, specs):
+    return _loadPlugins(
+        parser,
+        specs,
+        loadFeatureWriterFromString,
+        "Failed to load --feature-writer:\n  {}: {}",
+    )
 
 
 def _loadFilters(parser, specs):
-    filters = []
-    for s in specs:
-        if s == "None":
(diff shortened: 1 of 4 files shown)


anthrotype:
@moyogo how are you going to pass the list of custom filters down to ufo2ft? Don't we also need to add a `filters` parameter to the ufo2ft compile functions (like we have for featureWriters) as well as the preProcessor class?

anthrotype:
actually we may need _both_ a `preFilters` and `postFilters` parameter for the compile functions and preprocessor class. Right now when preprocessor calls `loadFilters(ufo)` it receives two lists, one for filters to be applied before the default filters (decomposition, remove overlap, etc.), and one list for filters to be run after default ones. The filter object itself does not know whether it is a pre or post kind of filter.

If we don't want to add two new parameters but only a single `filters` one, then instead of a flat list we could pass a dictionary with `filters={"pre": [...], "post": []}`, or a 2-tuple `filters=([...], [...])` listing both pre and post.

The command-line option currently is a generic `--filters` which doesn't communicate whether it's going to be a pre- or post-filter. I think also for the CLI, we may need two options `--pre-filters` and `--post-filters`.

moyogo:
> @moyogo how are you going to pass the list of custom filters down to ufo2ft? Don't we also need to add a filters parameter to the ufo2ft compile functions (like we have for featureWriters) as well as the preProcessor class?

Yes. I'll open a PR.

> If we don't want to add two new parameters but only a single filters one, then instead of a flat list we could pass a dictionary with filters={"pre": [...], "post": [...]}, or a 2-tuple filters=([...], [...]) listing both pre and post.

> The command-line option currently is a generic --filters which doesn't communicate whether it's going to be a pre- or post-filter. I think also for the CLI, we may need two options --pre-filters and --post-filters.

The [loadFilters()](https://github.com/googlefonts/ufo2ft/blob/97eb6734397dcb2bc395775c4e342d74cd6474a3/Lib/ufo2ft/filters/__init__.py#L53) function uses the `pre` argument for put filters in pre-filters or post-filters.
Can we add a `BaseFilter.pre` attribute that would be False by default?
BasePreProcessor could then sort them into preFilters and postFilters.
That way we can pass a list of filters (like in the UFO lib key) and only sort them out before we use them (like loadFilters currently does).

moyogo:
See https://github.com/googlefonts/ufo2ft/pull/479

anthrotype:
https://github.com/googlefonts/ufo2ft/releases/tag/v2.20.0

Lib/fontmake/__main__.py
```diff
@@ -40,6 +41,21 @@ def _loadFeatureWriters(parser, specs):
     return feature_writers
 
 
+def _loadFilters(parser, specs):
```
anthrotype, line 44: this _loadFilters looks exactly like _loadFeatureWriters, maybe parametrize them

anthrotype (commented):
a couple tests would be nice

anthrotype (approved):

PR #760, opened 2021-04-09
[setup.py] Work around pip bug with "pip install -e ."

behdad:
See https://github.com/pypa/pip/issues/7953

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,10 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import site
 import sys
 from io import open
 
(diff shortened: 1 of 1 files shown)


anthrotype (approved):
I don't use pip install -e --user together since I use virtual environments for development, but I don't see why this should not work. So LGTM as long as it doesn't break anything else

behdad:
Thanks.

behdad:
I don't have permission to merge.

anthrotype:
I sent invitation, try again afterwards and let me know otherwise

PR #779, opened 2021-06-14
Make "fontmake Thing.glyphs -o variable" work

simoncozens:
I don't know; we use these expensive, sophisticated computers, and then we force the users to tell them that a file ending `.designspace` is a designspace file. I'm forever forgetting to specify the `-g`/`-m`/`-u` option, and I'm sure part of the reason why I always forget to add it is because it's *totally redundant*. (Another reason is that in most Unix utilities, the file to work on is the default positional argument, so `fontmake InputFile.ufo` makes ergonomic sense.)

(Oh, and I always forget that designspace is specified using `-m`, not `-d`, *for historical reasons*.)

This PR interprets any left-over unparsable arguments as input file names and sorts them into the appropriate input file format.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -110,7 +110,7 @@ def main(args=None):
         title="Input arguments",
         description="The following arguments are mutually exclusive (pick only one):",
     )
-    xInputGroup = inputGroup.add_mutually_exclusive_group(required=True)
+    xInputGroup = inputGroup.add_mutually_exclusive_group()
     xInputGroup.add_argument(
         "-g", "--glyphs-path", metavar="GLYPHS", help="Path to .glyphs source file"
     )
@@ -425,7 +425,9 @@ def main(args=None):
         help="Configure the logger verbosity level. Choose between: "
         "%(choices)s. Default: INFO",
     )
-    args = vars(parser.parse_args(args))
+
+    parsed_args, unknown = parser.parse_known_args(args)
+    args = vars(parsed_args)
 
     specs = args.pop("feature_writer_specs")
     if specs is not None:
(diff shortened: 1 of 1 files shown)


behdad:
FINALLY!!!

anthrotype:
thanks! I refrained from changing this to avoid breaking existing workflows and shell scripts, but as long as the old invocations keep working after this patch, I'm happy to have this in

Lib/fontmake/__main__.py
```diff
@@ -438,6 +440,31 @@ def main(args=None):
...
+    for file in unknown:
+        if any_input and not (ufo_paths and file.endswith(".ufo")):
+            logging.warn(
+                "Only one input file type accepted; additional file '%s' ignored" % file
+            )
+        elif file.endswith(".glyphs"):
+            glyphs_path = file
+            any_input = True
+        elif file.endswith(".designspace"):
+            designspace_path = file
```
anthrotype, line 454: -g/-m/-u are mutually exclusive (enforced by argparse), whereas this seems to allow specifying e.g. _both_ e.g. a .glyphs and a .designspace file, I don't think we want to allow that in the same fontmake invocation. I suggest checking that mutual exclusivity of input type is still the case

anthrotype: oh I just realised what you're doing now, you are ignoring additional input file types and only taking the first one with a warning. I would probably let it blow up with an error instead, similarly to specifying both `-g` and `-m` in the same invokation

Lib/fontmake/__main__.py
```diff
@@ -438,6 +440,31 @@ def main(args=None):
...
+            if not ufo_paths:
+                ufo_paths = []
+            ufo_paths.append(file)
+            any_input = True
+        else:
+            logging.warning("Unknown input file type '%s' ignored" % file)
+    if not any_input:
+        parser.print_help(sys.stderr)
+        print("fontmake: No input files specified")
+        sys.exit(1)
```
anthrotype, line 466: you can use `parser.error()` here (see elsewhere in the file) instead of print + sys.exit

anthrotype:
the default `fontmake -h` as generated by argparse probably doesn't mention the fact that residual positional arguments are interpreted as inputs, not sure if/how we should advertise this feature somewhere

PR #810, opened 2021-09-27
Use glyphsLib minimal builds

khaledhosny:
This tells glyphsLib to create enough UFO for building binary fonts and discard any unneeded data.

This requires glyphsLib 6.0.0 or later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "pypy-3.7"]
+        python-version: [3.7, 3.8, 3.9] # "pypy-3.7", see #810
         platform: [ubuntu-latest, windows-latest]
         exclude:
           - platform: windows-latest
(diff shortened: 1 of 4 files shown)


anthrotype (approved):

khaledhosny:
Looks like pypy3 does not like the new glyphsLib parser.

anthrotype:
uh-oh yes.. https://github.com/fonttools/openstep-plist/issues/12

anthrotype:
I had not realised that with glyphsLib 6 and the concomitant switch to openstep-plist parser, we would lose pypy support. That's a pity, but for the time being there's no quick fix and we may have to temporarily drop support for pypy until we figure out how to make it happy. I suggest we open an issue here and/or on glyphsLib to track this, linking to https://github.com/fonttools/openstep-plist/issues/12, and here we comment-out the pypy3 tests in the CI config file (with a TODO and a link to the issue)

khaledhosny:
> we temporarily comment out the pypy3 tests in the CI config file (with a TODO and a link to the issue)

Done

anthrotype (approved):

anthrotype:
I think we can tag a major version bump v3.0.0b1 instead of v2.5.0b1 -- given we dropped support for python 3.6 and pypy3, we add support for Glyphs 3, we will soon. add support for iondrive (#802)

anthrotype:
@khaledhosny https://pypi.org/project/fontmake/3.0.0b1/

khaledhosny:
Thanks!

PR #832, opened 2021-12-09
Add compatibility checker

simoncozens:
Currently for [Radio Canada Italic](https://github.com/cbcrc/radiocanadafonts/commit/a94192148f8344eb343218c5290ce9aaecc45fc2), fontmake produces this output:

```
fontmake: Error: In 'sources/RadioCanada-Italic.glyphs' -> 'master_ufo/RadioCanada-Italic.designspace': Generating fonts from Designspace failed: Base master not found.
```

This is particularly confusing because earlier on in the log it says:

```
INFO:fontTools.varLib:Index of base master: 4
```

So why can't it find the base master? Not an easy question for users to answer. With this PR, fontmake says:

```
ERROR:fontmake.compatibility:Fonts had differing anchors in glyph cedi:
ERROR:fontmake.compatibility: * Radio Canada Condensed Light Italic had "top"
ERROR:fontmake.compatibility: * Radio Canada Condensed Italic, Radio Canada Condensed Bold Italic, Radio Canada Light Italic, Radio Canada Italic, Radio Canada Bold Italic had ""
fontmake: Error: In 'sources/RadioCanada-Italic.glyphs' -> 'master_ufo/RadioCanada-Italic.designspace': Compatibility check failed
```

The checker is applied when producing variable fonts or whether the Glyphs "Enforce Compatibility Check" custom parameter is turned on.


anthrotype:
Thanks Simon, this is super useful!! Mind adding a few tests as well?

simoncozens:
Will do, just as soon as I've got the ones we already have passing...

simoncozens:
All ready!

Lib/fontmake/font_project.py
```diff
@@ -951,7 +951,8 @@ def run_from_designspace(
             preFilters, postFilters = loadFilters(designspace)
             filters = preFilters + postFilters
 
-        source_fonts = [source.font for source in designspace.sources]
+        designspace_copy = self._load_designspace_sources(designspace)
```
anthrotype, line 954: I wonder why we need to load designspace sources twice..

tests/test_main.py
```diff
@@ -76,9 +76,10 @@ def test_interpolation_mutatormath(data_dir, tmp_path):
 
 
 def test_interpolation_mutatormath_source_layer(data_dir, tmp_path):
-    shutil.copyfile(
-        data_dir / "MutatorSans" / "MutatorSans.designspace",
-        tmp_path / "MutatorSans.designspace",
+    shutil.copytree(
+        data_dir / "MutatorSans",
```
anthrotype: why did you need to copy the sources too?

Lib/fontmake/font_project.py
```diff
@@ -949,6 +951,12 @@ def run_from_designspace(
             preFilters, postFilters = loadFilters(designspace)
             filters = preFilters + postFilters
 
+        designspace_copy = self._load_designspace_sources(designspace)
+        source_fonts = [source.font for source in designspace_copy.sources]
+        if interp_outputs or any(COMPAT_CHECK_KEY in font.lib for font in source_fonts):
```
anthrotype, line 956: should you simply check if COMPAT_CHECK_KEY is in the designspace.lib instead of any(font.lib)?

tests/test_main.py
simoncozens: Because otherwise this test will fail out because of the compatibility checker not being able to find the sources; we need to get past that in order to test for the "layer in designspace" thing we're testing for in this particular test.

Lib/fontmake/font_project.py
simoncozens: The "interpolate layout" check got upset about something. I don't really understand it: `font["GPOS"] = gpos_src["GPOS"]` died because `gpos_src` didn't have a "GPOS" key.

Lib/fontmake/font_project.py
simoncozens: That key doesn't get copied to the designspace.

Lib/fontmake/font_project.py
```diff
@@ -54,6 +55,7 @@
     GLYPHS_PREFIX + "customParameter.InstanceDescriptorAsGSInstance.Keep Glyphs"
 )
 GLYPH_EXPORT_KEY = GLYPHS_PREFIX + "Glyphs.Export"
+COMPAT_CHECK_KEY = GLYPHS_PREFIX + "customParameter.GSFont.Enforce Compatibility Check"
```
anthrotype, line 58: how about we also add a CLI option like --compatibility-check or something? I imagine a DS+UFO workflow could take advantage of this as well, and that priavet lib key is pretty obscure

Lib/fontmake/font_project.py
anthrotype: Yeah, in fact I'm suggesting that it should. Who's responsible for it? glyphsLib I imagine

anthrotype: So, the Glyphs user sets that parameter on the GSFont or on the GSMaster? If it's on the font, then it's global and should belong to the DS.lib, if specific to indiviodual masters, then it's ok in font.lib but then should you not only include in the compat check the font masters that actually have the key? Currently you check all if any one has the key, IIUC.

simoncozens: It's set on the font, so it should go to the DS lib. But we need the sources anyway to check them, so there's not much in it.

anthrotype: I still think it would be nicer if we checked the DS.lib only, since this is a global parameter that applies to all masters, not some of them. The any() check followed by compat check for all is confusing to me.

Lib/fontmake/font_project.py
anthrotype: i'd prefer we get to the bottom of this and try to avoid the reload; if we can't we should add a comment explaining why we do that

Lib/fontmake/font_project.py
anthrotype, line 58: How about we also add an --check-compatibilty CLI flag (False by default, ignored for interpolatable outputs [always True by definition for those])? That'd make it easier to use from non-Glyphs.app workflows, as that lib key is pretty long and obscure.

Lib/fontmake/font_project.py
anthrotype: ok fine, one day we may fix glyphsLib to export global font parameters to the designspace.lib. For now I suggest you add a comment that says something like "glyphsLib currently stores this global GSFont parameter in all the master UFO's lib so we check if it's set in any of them"

Lib/fontmake/font_project.py
simoncozens: Got it. The interpolate-layout stuff wasn't loading up the binary fonts for the `interpolate_layout_from` designspace because it already had (UFO) fonts in place.

anthrotype: Thanks!

Lib/fontmake/font_project.py
simoncozens: Actually this solves the above problem too. If we create a `check_compatibility` option then we can fill it from the Glyphs file instead of grubbing around in lib files.

anthrotype: > we can fill it from the Glyphs file

sure but how do you fill a CLI option from the Glyphs file exactly?

Lib/fontmake/font_project.py
```diff
@@ -172,6 +180,9 @@ def build_master_ufos(
         except Exception as e:
             raise FontmakeError("Loading Glyphs file failed", glyphs_path) from e
 
+        if font.customParameters["Enforce Compatibility Check"]:
+            self.check_compatibility = True
```
anthrotype, line 184: hm the FontProject object is stateful.. It's fine if the CLI option is set on the object self.check_compatibility attribute, since the FontProject is made there and thrown away after main() is done. But the fact that one particular GSFont has that customparameter should not change the state of the FontProject's check_compatibility option (which would apply for subsequent runs as well). Could you find a way to trickle that option down to where it's used without modifying self?

Lib/fontmake/font_project.py
```diff
@@ -172,6 +180,9 @@ def build_master_ufos(
         except Exception as e:
             raise FontmakeError("Loading Glyphs file failed", glyphs_path) from e
 
+        if font.customParameters["Enforce Compatibility Check"]:
```
anthrotype, line 183: traditionally, when there are Glyphs-only custom parameters involved, given that the pipeline works with UFOs internally, we have looked those up inside the UFO or Designspace libs (as exported by glyphsLib from the original .glyphs source) instead of directly in the GSFont instance. This makes it possible for DS+UFO workflows to avail themselves of these, or to build from .glyphs source in stages (first export to DS+UFO, then maybe do something else, then continue on to build off that).

If you check the GSFont custom parameter directly, it's not possible for a DS worfklow to store this option anywhere, requiring one to pass the CLI option.

Ideally we'd have our own (com.github.googlefonts.fontmake..) lib key that stores that in the DS.lib and glyphsLib writes that out. Maybe we can add later as needed

simoncozens: Oh man. This is like playing Jenga. :-)

anthrotype: welcome to my life 😅

anthrotype: I think you also need to add `check_compatibility` to `exclude_args` in here, just before running pipeline from single UFOs:

https://github.com/googlefonts/fontmake/blob/f36543baf5065f05e7fb0c5ea58d7eab8bc9255a/Lib/fontmake/__main__.py#L572-L584

Lib/fontmake/font_project.py
```diff
@@ -419,6 +419,7 @@ def save_otfs(
         flatten_components=False,
         filters=None,
         generate_GDEF=True,
+        check_compatibility=None,
```
anthrotype, line 422: sorry no, please see my comment https://github.com/googlefonts/fontmake/pull/832#discussion_r765778028

anthrotype (approved):
LGTM, thanks for the patience!

anthrotype:
maybe squash before you merge

simoncozens:
Thanks for your patience!

JeremieHornus:
We have a lot of new issues with this compatibility checker that seem to be bugs in here.
It spots glyphs with "different number of components" where there are actually no components at all :( 
please advise

JeremieHornus:
ah sorry! it was actual compatibility problems in our UFO sources! hard to spot but real issues
so thanks a lot for this 'compatibility check' addition 💯

simoncozens:
I have a problem with it, though - it's doing compatibility checks on non-exported glyphs, and it doesn't matter if they aren't compatible.

justvanrossum:
It seems it also complain about anchor order, say "top, bottom" vs. "bottom, top". Shouldn't a case like that "just work"? Or are there cases where the anchor order is meaningful?

simoncozens:
No, that should just work. I'll sort it.

Issue #562, opened 2019-06-17
fontmake should use ttfautohint-py wrapper


anthrotype:
https://github.com/fonttools/ttfautohint-py

and should remove that `fontmake/ttfautohint.py` module

hyvyys:
Not sure if related, I get this

```'../../../../../../tmp/tmprxum4866.ttf': Could not launch ttfautohint (is it installed?): [Errno 2] No such file or directory: 'ttfautohint': 'ttfautohint'```

And fontmake is unable to hint the fonts.

PR #850, opened 2022-01-28
Chain autohinting parameters from Glyphs source to fontmake

yanone:
I wouldn't know how to write a test for this, so unless you insist, I'm skipping

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -546,11 +546,8 @@ class FontProject:
                     font["GSUB"] = gsub_src["GSUB"]
 
             # Decide on autohinting and its parameters
-            autohint_thisfont = (
-                ttf and (
-                    autohint
-                    or ufo.lib.get(AUTOHINTING_PARAMETERS)
-                )
+            autohint_thisfont = ttf and (
+                autohint or ufo.lib.get(AUTOHINTING_PARAMETERS)
             )
 
             if autohint_thisfont:
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -61,6 +61,7 @@ INTERPOLATABLE_OUTPUTS = frozenset(
     ["ttf-interpolatable", "otf-interpolatable", "variable", "variable-cff2"]
 )
 
+AUTOHINTING_PARAMETERS = GLYPHS_PREFIX + "customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"
 
 @contextmanager
 def temporarily_disabling_axis_maps(designspace_path):
@@ -541,14 +542,8 @@ class FontProject:
                     font["GSUB"] = gsub_src["GSUB"]
 
             # Read autohinting parameters from ufo lib if present
-            if (
-                "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"
-                in ufo.lib
-                and autohint is None
-            ):
-                autohint = ufo.lib[
-                    "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"
-                ]
+            if AUTOHINTING_PARAMETERS in ufo.lib and autohint is None:
+                autohint = ufo.lib[AUTOHINTING_PARAMETERS]
 
             do_autohint = ttf and autohint is not None
 
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -427,7 +427,8 @@ class FontProject:
             ttf: If True, build fonts with TrueType outlines and .ttf extension.
             is_instance: If output fonts are instances, for generating paths.
             autohint: Parameters to provide to ttfautohint. If not provided, the
-                autohinting step is skipped.
+                UFO lib is scanned for autohinting parameters. If nothing is found, 
+                the autohinting step is skipped.
             subset: Whether to subset the output according to data in the UFOs.
                 If not provided, also determined by flags in the UFOs.
             use_production_names: Whether to use production glyph names in the
@@ -523,8 +524,6 @@ class FontProject:
             inplace=True,  # avoid extra copy
         )
 
-        do_autohint = ttf and autohint is not None
-
         for font, ufo in zip(fonts, ufos):
             if interpolate_layout_from is not None:
                 master_locations, instance_locations = self._designspace_locations(
@@ -540,6 +539,12 @@ class FontProject:
                     font["GDEF"] = gsub_src["GDEF"]
                 if "GSUB" in gsub_src:
                     font["GSUB"] = gsub_src["GSUB"]
+
+            # Read autohinting parameters from ufo lib if present
+            if "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options" in ufo.lib and autohint is None:
+                autohint = ufo.lib["com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"]
+
+            do_autohint = ttf and autohint is not None
 
             if do_autohint:
                 # if we are autohinting, we save the unhinted font to a
@@ -577,6 +582,7 @@ class FontProject:
                     ufo, ext, is_instance, autohinted=True, output_dir=output_dir
                 )
             try:
+                logger.info("Autohinting %s", otf_path)
                 ttfautohint(otf_path, hinted_otf_path, args=autohint)
             except TTFAError:
                 # copy unhinted font to destination before re-raising error
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -475,7 +475,7 @@ class FontProject:
                 pre-filters or post-filters, called before or after the default
                 filters. The default filters are format specific and some can
                 be disabled with other arguments.
-        """
+        """  # noqa: B950
         assert not (output_path and output_dir), "mutually exclusive args"
 
         if output_path is not None and len(ufos) > 1:
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -541,8 +541,14 @@ class FontProject:
                     font["GSUB"] = gsub_src["GSUB"]
 
             # Read autohinting parameters from ufo lib if present
-            if "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options" in ufo.lib and autohint is None:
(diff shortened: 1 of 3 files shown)


Lib/fontmake/font_project.py
```diff
@@ -541,6 +540,12 @@ def save_otfs(
                 if "GSUB" in gsub_src:
                     font["GSUB"] = gsub_src["GSUB"]
 
+            # Read autohinting parameters from ufo lib if present
+            if "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options" in ufo.lib and autohint is None:
+                autohint = ufo.lib["com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"]
```
anthrotype: you should move this lib key name at the top of the module next to similar constants.
It's pretty horrible name by the way, but I suppose that's what glyphsLib spits out.. /cc @belluzj

Lib/fontmake/font_project.py
```diff
@@ -541,6 +540,18 @@ def save_otfs(
...
                     font["GSUB"] = gsub_src["GSUB"]
 
+            # Read autohinting parameters from ufo lib if present
+            if (
+                "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"
+                in ufo.lib
+                and autohint is None
+            ):
+                autohint = ufo.lib[
+                    "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options"
```
anthrotype, line 550: let's keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

Lib/fontmake/font_project.py
```diff
@@ -427,7 +427,8 @@ def save_otfs(
             ttf: If True, build fonts with TrueType outlines and .ttf extension.
             is_instance: If output fonts are instances, for generating paths.
             autohint: Parameters to provide to ttfautohint. If not provided, the
-                autohinting step is skipped.
+                UFO lib is scanned for autohinting parameters. If nothing is found,
```
anthrotype, line 430: maybe It's also worth mentioning the lib key name here in the --help

Lib/fontmake/font_project.py
simoncozens: I would put this into a separate variable, because here you are pulling out a per-master setting and using it to override the settings for the whole project. If the next master in the loop doesn't have autohint options,  `autohint` will contain the settings from the previous master.

Lib/fontmake/font_project.py
```diff
@@ -541,6 +540,12 @@ def save_otfs(
                 if "GSUB" in gsub_src:
                     font["GSUB"] = gsub_src["GSUB"]
 
+            # Read autohinting parameters from ufo lib if present
+            if "com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options" in ufo.lib and autohint is None:
```
simoncozens: Please put this long string as a constant at the top of the file, similar to `KEEP_GLYPHS_NEW_KEY`

simoncozens (changes requested):

anthrotype:
> I wouldn't know how to write a test for this, so unless you insist, I'm skipping

I would insist, if you guys don't mind :)
You can make a simple test .glyphs file with one master and one instace, add the custom parameter to the instance, check that in the tests directory along side other test files, then add a new test test_main.py where you call fontmake to create the instance ttf, and then load the built TTFont and make sure that it's hinted (check for some key hinting table like fpgm or whatever).

anthrotype:
to make the test run on the CI, you would have to first install ttfautohint. The easiest would be to use homebrew on the mac CI runner.
This reminds me we have ttfautohint-py (which is pip installable) but fontmake isn't making use of it.. https://github.com/googlefonts/fontmake/issues/562

Lib/fontmake/font_project.py
yanone: I don't understand the hint. Is this the same remark as Simon’s regarding re-using the `autohint` variable?

anthrotype: yes, we don't want to repeat ourselves, so move the constant at the top, all good now

Lib/fontmake/font_project.py
yanone: The actual complete Glyphs key `com.schriftgestaltung.customParameter.InstanceDescriptorAsGSInstance.TTFAutohint options`?

Lib/fontmake/font_project.py
```diff
@@ -577,6 +589,7 @@ def save_otfs(
                     ufo, ext, is_instance, autohinted=True, output_dir=output_dir
                 )
             try:
+                logger.info("Autohinting %s", otf_path)
                 ttfautohint(otf_path, hinted_otf_path, args=autohint)
```
simoncozens, line 593: autohint_thisfont

Lib/fontmake/font_project.py
anthrotype: yes.. the alternative is one has to look at the code to find that out, since it's pretty obscure

Lib/fontmake/font_project.py
anthrotype: that is why one needs tests 👍

yanone: Please be specific. I don't understand the issue here

Lib/fontmake/font_project.py
yanone: Like this?: https://github.com/googlefonts/fontmake/pull/850/commits/b23d2291803c9f77b3d07577bcf864c79f96d886

Lib/fontmake/font_project.py
```diff
@@ -541,6 +541,18 @@ def save_otfs(
...
+            # Read autohinting parameters from autohint variable
+            if autohint is not None:
+                autohint_thisfont = autohint
+            # otherwise read autohinting parameters from ufo lib if present
+            elif AUTOHINTING_PARAMETERS in ufo.lib and autohint is None:
+                autohint_thisfont = ufo.lib[AUTOHINTING_PARAMETERS]
+            else:
+                autohint_thisfont = None
+
+            do_autohint = ttf and autohint_thisfont is not None
```
anthrotype, line 558: you could probably simplify this whole logic like so

```python
autohint_thisfont = (
    ttf and (
        autohint
        or ufo.lib.get(AUTOHINTING_PARAMETERS)
    )
)

if autohint_thisfont:
    ...
```

Lib/fontmake/font_project.py
anthrotype: yeah.. though the `\\` at the line end may be confusing, i'd remove that and keep on a single line (even if it may look too long or if it wraps to newline), easier to copypaste

Lib/fontmake/font_project.py
yanone: Oh, I get it. Simon marked the wrong line. the ttfautohint() needs the new variable instead.

anthrotype: sorry. You are not using the `autohint_thisfont` local variable that you just made exactly where you need it most: i.e. to the args parameter of ttfautohint. It is still using the global `autohint` one

Lib/fontmake/font_project.py
yanone: Not really, because in your suggestion `autohint_thisfont` is a boolean, while in my code it contains the autohinting parameters, which I would then still need in a separate variable. My code is analog the old code which also has two separate variables, except that `autohint_thisfont` is now font-specific

anthrotype: it's not only a boolean, it can be True, False, None or some string argument.. so technically a `Union[bool, str, NoneType]`. Try it your REPL.

yanone: I renamed the variable to `autohint_parameters` to be clear

yanone: Oh. I didn't know that adding a string to a boolean becomes the string or `False`. Okay.

anthrotype: https://realpython.com/python-and-operator/#using-pythons-and-operator-with-common-objects

yanone: Ok, done. Please sign off on this file before I will attempt a test.

anthrotype: yes, it's looking ok thanks

yanone:
> to make the test run on the CI, you would have to first install ttfautohint. The easiest would be to use homebrew on the mac CI runner.

Please advise on how to do this. I don't get the CI setup here.

yanone:
I added a test. Please review it. I'm now quietly assuming that TTFAutohint is installed via brew.

yanone:
> I'm now quietly assuming that TTFAutohint is installed via brew.

Of course it fails now. Waiting for further instructions

anthrotype:
After https://github.com/googlefonts/fontmake/pull/851 is merged on main, you should be able to merge main into your PR branch (or rebase on main) and your test should just work

anthrotype:
@yanone I merged #851, so when you can please git pull, then git merge main and push.

yanone:
At the expense of exposing my ignorance, I can't get this to run. Neither of your proposed commands do anything in my fork, be it main branch or the PR's branch. I'm not good with working with branches. I normally update my fork with `git merge upstream/main`, but that also doesn't work: `Already up to date.`

anthrotype:
yes, that's what I mean `git merge upstream/main`. But first you need to `git fetch upstream` otherwise you don't see the newly added commits

yanone:
So that line with the lib key is too long for flake8, which is why I originally broke it.

anthrotype:
Add `# noqa: B950` next to it to mute flake8

yanone:
Ands the pypy test fails (to install ttfautohint-py): https://github.com/googlefonts/fontmake/runs/4982698831?check_suite_focus=true#step:5:510

anthrotype:
> And the pypy test fails

pls ignore that for now, I'll investigate and fix that separately.

The PR looks good to me, I'll merge and take over from here. Thank you!

yanone:
Thanks

yanone:
It's OFL, so I didn't care. I already subsetted it to just the basic Latin A-Z and a-z. First I had just the A in it, but ttfautohint required a bunch of different characters to take measurements from, so I re-added the basics. I would have to dive into ttfautohint to make the minimum subset.

anthrotype:
I filed https://github.com/fonttools/ttfautohint-py/issues/11 for ttfautohint-py failing on pypy3. I'll just skip the test on pypy for now.

PR #855, opened 2022-02-11
CompatibilityChecker: make large error messages more concise

### LINKED ISSUE #854: CompatibilityChecker: improve output
I'm currently working on RobotoFlex which contains 71 masters. If I delete a single anchor, I get the following output:

Screenshot 2022-02-11 at 10 48 39

If you had 5+ missing anchors, you'd quickly experience fatigue (I'm currently at that phase).

Perhaps it would be nicer to instead have:

```
Fonts had differing anchors in glyph A:
* 70 fonts had "bottom, ogonek, top"
* RobotoFlex opsz8-wght400-wdth100-slnt-1 had "bottom, ogonek"
```

@simoncozens thoughts?

m4rc1e:
If there's more than 6 fonts, only show the font count.

Before

```
Fonts had differing anchors in glyph A:
 * RobotoFlex GRAD-200 had "bottom, ogonek"
 * RobotoFlex GRAD150, RobotoFlex opsz144, RobotoFlex opsz8, 
RobotoFlex wdth151, RobotoFlex wdth25, RobotoFlex wght100, 
RobotoFlex wght 1000, RobotoFlex Regular had "bottom, ogonek, top"
```

After

```
Fonts had differing anchors in glyph A:
 * RobotoFlex GRAD-200 had "bottom, ogonek"
 * 8 fonts had "bottom, ogonek, top"
```

Fixes #854

### DIFF
--- a/Lib/fontmake/compatibility.py
+++ b/Lib/fontmake/compatibility.py
@@ -69,9 +69,13 @@ class CompatibilityChecker:
             logger.debug(f"All fonts had same {what} in {context}")
             return True
         logger.error(f"Fonts had differing {what} in {context}:")
+        debug_enabled = logger.isEnabledFor(logging.DEBUG)
         for value, fonts in values.items():
-            fontlist = ", ".join(fonts)
-            logger.error(f" * {fontlist} had {value}")
+            if debug_enabled or len(fonts) <= 6:
+                key = ", ".join(fonts)
+            else:
+                key = f"{len(fonts)} fonts"
+            logger.error(f" * {key} had {value}")
         self.okay = False
         return False
 


m4rc1e:
Sorry, I didn't explain why I made this pr. I'm currently dealing with a family which has 71 masters. If there's a single incompatibility, the output is pretty much incomprehensible since it prints every family name for each missing anchor etc (could be hundreds)

Perhaps I could improve this pr by showing the full input if fontmake is run in debug mode, otherwise just show the concise message in this pr?

anthrotype:
sorry I missed this one. Yeah, I think you could print the full list of font names in debug mode and the shortened one in normal mode

anthrotype:
you can use `logger.isEnabledFor(logging.DEBUG)` => https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor to check for that

m4rc1e:
Thanks Cosimo, your logging approach works nicely!

The CompatibilityChecker will now run in verbose mode if the user runs fontmake with `--verbose DEBUG`. This is the current behaviour

```
(venv) Marcs-Air-2:fontmake marcfoley$ fontmake -m ~/Type/upstream_repos/Roboto-Flex/sources/RobotoFlex.designspace -o variable --verbose DEBUG
DEBUG:fontmake.font_project:ufoLib UFO validation is disabled
DEBUG:fontmake.compatibility:All fonts had same number of contours in glyph .notdef
DEBUG:fontmake.compatibility:All fonts had same number of points in glyph .notdef, contour 0
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 0, point 0
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 0, point 1
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 0, point 2
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 0, point 3
DEBUG:fontmake.compatibility:All fonts had same number of points in glyph .notdef, contour 1
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 1, point 0
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 1, point 1
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 1, point 2
DEBUG:fontmake.compatibility:All fonts had same point type in glyph .notdef, contour 1, point 3
DEBUG:fontmake.compatibility:All fonts had same number of points in glyph .notdef, contour 2
...
```

If debug isn't enabled (new default behaviour), for the above we'll get:

```
(venv) Marcs-Air-2:fontmake marcfoley$ fontmake -m ~/Type/upstream_repos/Roboto-Flex/sources/RobotoFlex.designspace -o variable
ERROR:fontmake.compatibility:Fonts had differing number of contours in glyph A:
ERROR:fontmake.compatibility: * 79 fonts had 2
ERROR:fontmake.compatibility: * RobotoFlex opsz14-wght100-wdth25 had 1
```

m4rc1e:
Let me tweak this a little.

Lib/fontmake/compatibility.py
```diff
@@ -16,11 +16,12 @@ def __exit__(self, type, value, traceback):
 
 
 class CompatibilityChecker:
-    def __init__(self, fonts):
+    def __init__(self, fonts, verbose=False):
         self.errors = []
         self.context = []
         self.okay = True
         self.fonts = fonts
+        self.verbose = verbose
```
anthrotype, line 24: you don't want to store the verbose attribute in the instance. Loggers have a global state. You just need to import the top-level fontmake.logger and check if it's enabled for DEBUG

anthrotype: the logger you want is actually fontmake.font_project.logger
https://github.com/googlefonts/fontmake/blob/0d48ff19b72dc0beed0c3835ea715392554e6dcf/Lib/fontmake/font_project.py#L46

anthrotype: actually, sorry, the fontmake.compatibility module has its own logger at 
https://github.com/googlefonts/fontmake/blob/0d48ff19b72dc0beed0c3835ea715392554e6dcf/Lib/fontmake/compatibility.py#L3

use that one. It inherits the logging handlers of the parent, so if the parent is enabled for logging, and itself hasn't been tweaked specially, then it will also be verbose.

m4rc1e: Much nicer! thank you.

Lib/fontmake/compatibility.py
```diff
@@ -71,7 +70,7 @@ def ensure_all_same(self, func, objs, what):
             return True
         logger.error(f"Fonts had differing {what} in {context}:")
         for value, fonts in values.items():
-            if self.verbose or len(fonts) <= 6:
+            if logger.isEnabledFor(logging.DEBUG) or len(fonts) <= 6:
```
anthrotype, line 73: move this `logger.isEnabledFor(logging.DEBUG)` check out of the for loop please, it's constant throughout the loop. Save it to a local variable and use in the loop

m4rc1e:
Will squash this once you're happy since it's a minor change with a load of commits.

anthrotype (approved):

m4rc1e:
Thanks Cosimo. Since it's a 6 liner, I've squashed everything into the first commit.

PR #867, opened 2022-03-16
Build DS5

belluzj:
This PR implements building a DesignSpace version 5 using fontmake. The main new feature is that a single input designspace can build several variable fonts.

This PR adds a command-line argument to fontmake to select which of the variable fonts to build, in the same way as one can filter which instances to build.

This PR needs:
* https://github.com/googlefonts/ufo2ft/pull/598
* https://github.com/fonttools/fonttools/pull/2436

The commit names are useless and I'll squash everything later, sorry about that.

@anthrotype


Lib/fontmake/font_project.py
```diff
@@ -306,9 +308,10 @@ def build_interpolatable_otfs(self, designspace, **kwargs):
         """
         return self._build_interpolatable_masters(designspace, ttf=False, **kwargs)
 
-    def build_variable_font(
+    def build_variable_fonts(
         self,
-        designspace,
+        designspace: designspaceLib.DesignSpaceDocument,
+        variable_fonts=True,
```
anthrotype, line 314: the type of this should be `str` and should default to `".*"` regex pattern (build all VFs regardless of the name); a bool value made sense for the instances parameter which one can set to False; setting `variable_fonts=False` in a method named `build_variable_fonts` doesn't make much sense.

anthrotype (approved):

PR #900, opened 2022-06-24
Add support for UFOZ as input format when no -u flag is used

### LINKED ISSUE #887: Add support for .ufoz (UFO compressed ZIP format)
According to [UFO 3 Specification](https://unifiedfontobject.org/versions/ufo3/), a UFO folder may be compressed into a ZIP format file with the extension .ufoz. `fontmake` currently does not support it with the following error: `fontmake: error: Unknown input file extension: xxx.ufoz` 

This will be a very good addition as CJK fonts that need to go up on Google Fonts has tens of thousands of glyphs that uploading the UFO folder is painstakeningly slow (without resorting to using Mac and Glyphs).

NightFurySL2001:
Resolves #887 .

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -131,7 +131,9 @@ def parse_mutually_exclusive_inputs(parser, args):
             if designspace_path:
                 parser.error("Only one *.designspace source file is allowed")
             designspace_path = filename
-        elif (filename.endswith(".ufo") and os.path.isdir(filename)) or (filename.endswith(".ufoz") and os.path.isfile(filename)) :
+        elif (filename.endswith(".ufo") and os.path.isdir(filename)) or (
+            filename.endswith(".ufoz") and os.path.isfile(filename)
+        ):
             ufo_paths.append(filename)
         else:
             parser.error(f"Unknown input file extension: '{filename}'")
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -13,8 +13,8 @@
 # limitations under the License.
 
 import logging
-import sys
 import os
+import sys
 from argparse import ArgumentParser, FileType
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


NightFurySL2001:
Sorry, I am quite new to contributing and am not sure on how to add tests 😅

anthrotype:
no worries.

can you please fix the linter errors here:
https://github.com/googlefonts/fontmake/runs/7042616527?check_suite_focus=true

the autoformatter (black) is complaining about line length or something.
You can install black and run on the modified files and the linte error should go away

anthrotype:
out of curiosity, does it work if you run `fontmake -u MyFont.ufoz` (with the `-u` flag) instead of without `-u` (implicit format guessed from positional arguments' file extension)? I thought that would already work.

I'm still ok to accept a patch like yours, just wanted to confirm that.

NightFurySL2001:
Yeah, using `-u` did work. It's just that it's not that easy to use lol
I'll fix the import too.

NightFurySL2001:
Also just a suggestion, might be better to provide the option to export `.ufoz` in future too. It's quite a pain in butt to export individual UFO folders for CJK fonts when one of them can easily surpass 10k+ glyphs. (and interpolating variable fonts... urgh)

anthrotype:
> might be better to provide the option to export .ufoz

yeah, sounds good. Would you like to work on that in this same PR?

NightFurySL2001:
Not really, I have no idea on how to do this...

anthrotype:
ok that's fine. We can keep #887 open until that's done. So this PR here only adds support for reading .ufoz when passed in to the fontmake CLI without the explicit -u flag.
What's left to do for #887 is also allow `ufoz` as an additional output format for `-o` option. That would only be relevant for .glyphs inputs, of course (i.e. if one wants to convert a .glyphs file to a Designspace file + a set of .ufoz masters).

PR #901, opened 2022-06-24
Set up testing for CPython 3.11

madig:


anthrotype:
it's still beta I think. Also I think we'll need to build the binary wheels for the dependencies. But if it's not stable it's not worth publishing those as the ABI may break still before final release, I think.

anthrotype:
> we'll need to build the binary wheels for the dependencies

below is the list of the ones we'll need to install fontmake without a local c/c++ compiler toolchain; marked are those which already have 3.11 wheels

- [ ] compreffor
- [ ] cu2qu
- [ ] lxml
- [ ] openstep-plist
- [ ] pyclipper
- [x] skia-pathops
- [x] unicodedata2

PR #905, opened 2022-07-01
Add options and save to .ufoz

### LINKED ISSUE #887: Add support for .ufoz (UFO compressed ZIP format)

NightFurySL2001:
Referring to both ufoLib and defcon, add ` structure={"package", "zip"}` to `save_ufo_as` function and appropriate calls.

Fixes #887 .

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -360,6 +360,15 @@ def main(args=None):
         action="store_false",
         help="Do not auto-generate a GDEF table, but keep an existing one intact.",
     )
+    outputGroup.add_argument(
+        "--save-ufo-as-zip",
+        dest="ufo_structure",
+        action="store_const",
+        const="zip",
+        default="package",
+        help="Save UFOs as .ufoz format. Only valid when generating UFO masters "
+        "from designspace or interpolating UFO instances.",
+    )
 
     contourGroup = parser.add_argument_group(title="Handling of contours")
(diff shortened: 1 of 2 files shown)


Lib/fontmake/__main__.py
```diff
@@ -360,6 +360,15 @@ def main(args=None):
         action="store_false",
         help="Do not auto-generate a GDEF table, but keep an existing one intact.",
     )
+    outputGroup.add_argument(
+        "--save-ufo-as-zip",
+        dest="ufo_structure",
+        action="store_const",
+        const="zip",
+        default="package",
+        help="Save UFOs as .ufoz format. Only valid when generating UFO masters "
```
anthrotype, line 369: actually one does not "generate UFO masters from designspace", they are the sources themselves. Maybe you meant "from .glyphs"

anthrotype: i'll change the wording myself, thanks for the PR

PR #916, opened 2022-07-19
Allow disabling the compatibility checker

simoncozens:
The compatibility checker is really nice for reporting errors but it's actually pretty heavy to run on fonts where we already have built it before and know things are OK. This saves a few seconds.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -335,7 +335,8 @@ def main(args=None):
         help="Enable ufoLib validation on reading/writing UFO files. It is "
         "disabled by default",
     )
-    outputGroup.add_argument(
+    compatibilityGroup = outputGroup.add_mutually_exclusive_group()
+    compatibilityGroup.add_argument(
         "--check-compatibility",
         action="store_true",
         help="Check if the source files are interpolatable. It is "
@@ -343,6 +344,13 @@ def main(args=None):
         "or what the 'Enforce Compatibility Check' custom parameter is "
         "set on a Glyphs file",
     )
+    compatibilityGroup.add_argument(
+        "--no-check-compatibility",
+        action="store_true",
+        dest="disable_compatibility_check",
+        help="Turns off the compatibility checker even when building variable fonts",
+    )
+
     outputGroup.add_argument(
         "--expand-features-to-instances",
         action="store_true",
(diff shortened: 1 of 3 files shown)


Lib/fontmake/__main__.py
```diff
@@ -335,14 +335,22 @@ def main(args=None):
...
         action="store_true",
         help="Check if the source files are interpolatable. It is "
         "disabled by default, but enabled when building variable fonts "
         "or what the 'Enforce Compatibility Check' custom parameter is "
         "set on a Glyphs file",
     )
+    compatibilityGroup.add_argument(
+        "--no-check-compatibility",
+        action="store_true",
+        dest="disable_compatibility_check",
```
anthrotype, line 350: how about you make the existing `check_compatibility` a 3-state option instead of adding a new `disable_compatibility_check` bool parameter: i.e. `None` by default, which means `True` for interpolable outputs, `False` otherwise; and then explicit `True` / `False`, respectively controlled by the two CLI flags --check-compatiblity / --no-check-compatibility; if omitted, default to None.

Lib/fontmake/__main__.py
```diff
@@ -346,8 +346,8 @@ def main(args=None):
     )
     compatibilityGroup.add_argument(
         "--no-check-compatibility",
-        action="store_true",
-        dest="disable_compatibility_check",
+        action="store_false",
+        dest="check_compatibility",
```
anthrotype, line 350: I think you also need to add default=None to both options, otherwise store_false/store_true will set the default to the opposite bool value, no?

anthrotype (commented):
a small test exercising the options would also be appreciated ;)

simoncozens:
Yep, working on a test.

anthrotype (approved):
LGTM. Maybe you could squash before merge to remove the git log noise about black, flake8, etc.

gusbemacbe:
The `--no-check-compatibility` seems to be ignored when I ran the command `fontmake -m HexVF.designspace --no-check-compatibility -o variable --production-names --output-path HexVF.ttf`, and I still received a lot of compatibility errors.

anthrotype:
Do the errors come from the fontmake compatibility checker or from further down the build pipeline? It might not be possible to build the VF if the outlines are not compatible, that's what the checker is for..

PR #919, opened 2022-07-28
clean up logging configuration

anthrotype:
logging configuration is global, does not belong to a specific `fontmake.FontProject` instance, so it should only be done inside `main()`. This PR deprecates the `verbose` and `timing` parameters to FontProject constructor and moves the logging configuration to the CLI module (`fontmake.__main__`).
The old parameter continue to work but a deprecation warning is issued (category is actually `UserWarning` as `DeprecationWarning` proper are usually muted -- just to shout it out more louder)

It also enables logging configuration for a logger named "ufo2ft.timer" (which reports timing info inside ufo2ft) which doesn't exist yet but will soon (https://github.com/googlefonts/ufo2ft/pull/641)
This and fontmake's own timing logger are controlled by the fontmake `--timing` flag, independently from the `--verbose LEVEL` flag which control the root logger configuration.

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -194,6 +194,7 @@ def _configure_logging(level=None, timing=False):
         # timing-related DEBUG messages get a separate configuration so we
         # can enable them without needing to lower the global verbosity level
         configLogger(logger="fontmake.timer", level=logging.DEBUG, format=fmt)
+        configLogger(logger="ufo2ft.timer", level=logging.DEBUG, format=fmt)
 
 
 def main(args=None):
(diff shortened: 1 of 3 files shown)


anthrotype:
I think this is fairly uncontroversial so I'll go on and merge. We can decide to decorate more methods with `@timer` or add context-manager `with timer(...):` to blocks of code that we deem useful. @simoncozens is doing that inhttps://github.com/googlefonts/ufo2ft/pull/641 and fontmake (with this PR) is already ready to handle messages from the "ufo2ft.timer" logger.

simoncozens:
One important thing now (since we do not have hierarchical timers...) is that we need to be careful to avoid decorating an operation and a constituent part of the same operation, or else we lose the ability to add the timings together. What I mean is:

```
Took 5.00s to run MarkFeatureWriter
Took 7.00s to run KernFeatureWriter
Took 12.00s to run feature writers
```

Unless you teach your analysis code that `MarkFeatureWriter` and `KernFeatureWriter` are inside `feature writers`, then it will seem like that part of the run took 24 seconds.

PR #931, opened 2022-09-01
Support building from glyphspackages

simoncozens:
This requires the next version of glyphsLib (i.e. one in which `glyphsLib.load` can take a path name).


anthrotype (approved):
LGTM, pending next glyphsLib release, bonus point if you add a small test 
thanks!

PR #941, opened 2022-09-29
Add CLI to fontmake.instantiator

simoncozens:
This adds a CLI interface to the instantiator; useful for testing/debugging and also actually instantiating UFOs!

Can be used either with an instance name:

```
% python3 -m fontmake.instantiator --output-dir instance_ufo GlyphsUnitTestSans.designspace Light
Generated instance 'Light'.
Saved to 'instance_ufo/GlyphsUnitTestSans-Light.ufo'.
```

or with coordinates:

```
% python3 -m fontmake.instantiator --output-dir instance_ufo GlyphsUnitTestSans.designspace wght=523
Generated instance 'wght=523'.
Saved to 'instance_ufo/GlyphsUnitTestSans-wght523.ufo'.
```

### DIFF
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -780,3 +780,98 @@ class Variator:
             return copy.deepcopy(self.location_to_master[normalized_location_key])
 
         return self.model.interpolateFromMasters(normalized_location, self.masters)
+
+
+if __name__ == "__main__":
+    import argparse
+    import os
+    import sys
+
+    from fontTools.designspaceLib import DesignSpaceDocument, InstanceDescriptor
+
+    parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


anthrotype:
i worry that we may then have to support multiple ways to accomplish the same thing, that's all

simoncozens:
That does make sense. I was seeing instantiator as a self-contained bit of code, but of course within fontmake it can do almost all of this.

anthrotype:
I actually wanted to move that module out of fontmake (which I see just as the command line front-end) to ufo2ft (which is a library doing most of the building logic), but never got around doing that.

simoncozens:
Could we make it a separate project altogether? That way if we ever need to replace the instantiator with a different implementation it's all nicely compartmentalised...

anthrotype:
yeah probably better that way

PR #944, opened 2022-10-05
Update glyphsLib to gain support for glyphspackage

rsms:
[glyphsLib v6.1.0](https://github.com/googlefonts/glyphsLib/releases/tag/v6.1.0) adds support for loading .glyphspackage sources. Updating it for fontmake would broaden the use of fontmake and remove the need for preprocessing steps with e.g. [glyphspkg](https://github.com/jenskutilek/glyphspkg)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 fonttools[unicode,lxml,ufo]==4.34.0; platform_python_implementation == 'CPython'
 fonttools[unicode,ufo]==4.34.0; platform_python_implementation != 'CPython'
 cu2qu==1.6.7.post1
-glyphsLib==6.0.6
+glyphsLib==6.1.0
 ufo2ft==2.28.0
 MutatorMath==3.0.1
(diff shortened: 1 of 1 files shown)


anthrotype:
Thanks! I think you also need to bump the minimum required version in setup.py for glyphsLib, which will be used when one does `pip install fontmake`. The requirements.txt is only used for setting up our own test CI environment.

Are we sure this is the only change needed? I think there's also this line (or similar elsewhere, not 100% sure that's the only one):

https://github.com/googlefonts/fontmake/blob/8f18f7305929346cb299fe20f5758fd7154aa0df/Lib/fontmake/__main__.py#L127

anthrotype:
updated glyphsLib with https://github.com/googlefonts/fontmake/commit/f7032384961191ea0a7f0fa1506b572f89a2ae8b

anthrotype:
> Are we sure this is the only change needed?

#931 took care of those

PR #958, opened 2022-12-05
Implement Remove Glyphs custom feature

simoncozens:
See https://github.com/googlefonts/glyphsLib/issues/735

- Implement Remove Glyphs custom feature
- Tests for Remove Glyphs

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -717,8 +717,8 @@ class FontProject:
         designspace.write(designspace_path)
 
     def subset_otf_from_ufo(self, otf_path, ufo):
-        """Subset a font using "Keep Glyphs" custom parameter and export flags as set
-        by glyphsLib.
+        """Subset a font using "Keep Glyphs"/"Remove Glyphs" custom parameters,
+        and export flags as set by glyphsLib.
 
         "Export Glyphs" is currently not supported:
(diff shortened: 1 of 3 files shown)


tests/test_main.py
```diff
@@ -367,6 +367,33 @@ def test_subsetting(data_dir, tmp_path, write_skipexportglyphs):
             font = fontTools.ttLib.TTFont(font_path)
             assert font.getGlyphOrder() == [".notdef", "space", "A", "C"]
 
+    shutil.copyfile(data_dir / "TestSubset2.glyphs", tmp_path / "TestSubset2.glyphs")
```
anthrotype, line 370: better to move this to its own separate test function

Lib/fontmake/font_project.py
```diff
@@ -709,7 +720,7 @@ def subset_otf_from_ufo(self, otf_path, ufo):
         """Subset a font using "Keep Glyphs" custom parameter and export flags as set
```
anthrotype, line 720:
```suggestion
        """Subset a font using "Keep Glyphs"/"Remove Glyphs" custom parameters, and export flags as set
```

anthrotype (approved):
so "Remove Glyphs" is basically the reverse of "Keep Glyphs", I guess it makes sense to use the subsetter for both. LGTM, thanks!

PR #962, opened 2022-12-08
Use MathGlyph's "strict" flag

### LINKED ISSUE #961: fontmake -i destroys geometry when off curve points coincide with on curve points
The enclosed archive contains a .designspace file using two ufos. Running the "generate_instances.py" script writes a third master, based on the instance defined in the .designspace file.

The single glyph shape in those ufos have a cluster of points at each corner: an on-curve point, two off-curve point, another on-curve point, all at the same (x, y) position. (These could then be used to interpolate to a sharp corner with rounded-corner sources.)

The output of the script has all off-curve point stripped, so is no longer compatible with the sources. (The purpose here is to insert the generated ufo as a new source in the designspace.)

The fontmake instantiator uses `fontMath.MathGlyph` to do the interpolation, which messes with the topology of the shapes. In this case it filters out off-curve points that coincide with neighboring on-curve points.

`MathGlyph` recently got a "strict" flag to disable this behavior, and I think fontmake should use it.

That said, fontMath has a bug of its own here: https://github.com/robotools/fontMath/pull/305

[fontmake_test.zip](https://github.com/googlefonts/fontmake/files/10186163/fontmake_test.zip)

justvanrossum:
This fixes #961, and makes instantiator.py as strict as varLib when interpolating.

Depends on https://github.com/robotools/fontMath/pull/305, though, as fontMath doesn't properly propagate its "strict" flag down to copies.

So this PR should probably also update its fontMath version dependency, once a new fontMath is out.

### DIFF
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -603,7 +603,7 @@ def collect_glyph_masters(
             source.location, axis_bounds
         )
         locations_and_masters.append(
-            (normalized_location, fontMath.MathGlyph(source_glyph))
+            (normalized_location, fontMath.MathGlyph(source_glyph, strict=True))
         )
 
     # Filter out empty glyphs if the default glyph is not empty.
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ cu2qu==1.6.7.post1
 glyphsLib==6.1.0
 ufo2ft==2.29.0
 MutatorMath==3.0.1
-fontMath==0.9.2
+fontMath==0.9.3
 defcon[lxml]==0.10.2; platform_python_implementation == 'CPython'
 defcon==0.10.2; platform_python_implementation != 'CPython'
 booleanOperations==0.9.0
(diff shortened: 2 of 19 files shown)


anthrotype (approved):
LGTM, pending the other upstream change

anthrotype:
ideally we also add a test perhaps based on the one from #961

justvanrossum:
That failing lint check seems to demand passing the `strict` kw arg to builtin `zip`. This was introduced in 3.10, so as long as we need to support Python < 3.10, this lint check seems overly aggressive.

Not sure what to do about that in the context of this PR.

anthrotype:
for now, we need to mute that overly zealous flake8 warning with https://github.com/googlefonts/ufo2ft/pull/684

anthrotype:
i'm fine if you do that within this same PR, thanks!

anthrotype:
oh jeez, codecov is hunting us again here.. ignore it for now, I can fix it on main

justvanrossum:
Oh I just added a test! Sure, I'm fine with merging if you consider it fine.

anthrotype:
can you please rebase on main? I think I fixed both flake8 and codecov issues

anthrotype (approved):
test also LGTM

anthrotype:
just pushed v3.5.1 tag 👍

Issue #603, opened 2019-11-26
Determine glyphs from default master


madig:
Cosimo chat log
```
fontmake instantiator collects the list of glyph_names by unioning the font.keys of all the masters
it should rather take the default_font.keys() only
also, in collect_glyph_masters it silently skips if a glyph_name is not in source_layer, without checking if that source_layer is the default, in which case it should blow away
if a non-default master has a glyph which the default doesn't have, then that glyph should be ignored by definition
```

PR #980, opened 2023-02-16
Instantiator support for non-default layer sources

verbosus:
This PR adds the ability to build from non-default layers. If the default designspace source is set to a non-default layer, then its glyphset is used as reference for all the other sources—glyphs that are not on the layer are discarded. I believe this should also address issue #603

### DIFF
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -137,7 +137,7 @@ def parse_mutually_exclusive_inputs(parser, args):
         ):
             ufo_paths.append(filename)
         else:
-            parser.error("Unknown input file extension: {!r}".format(filename))
+            parser.error(f"Unknown input file extension: {filename!r}")
 
     count = sum(bool(p) for p in (glyphs_path, ufo_paths, designspace_path))
     if count == 0:
--- a/Lib/fontmake/__main__.py
+++ b/Lib/fontmake/__main__.py
@@ -137,7 +137,7 @@ def parse_mutually_exclusive_inputs(parser, args):
         ):
             ufo_paths.append(filename)
         else:
-            parser.error(f"Unknown input file extension: '{filename}'")
+            parser.error("Unknown input file extension: {!r}".format(filename))
 
     count = sum(bool(p) for p in (glyphs_path, ufo_paths, designspace_path))
     if count == 0:
--- a/Lib/fontmake/errors.py
+++ b/Lib/fontmake/errors.py
@@ -26,7 +26,7 @@ class FontmakeError(Exception):
 
     def __str__(self):
         trail = " -> ".join(
-            "{!r}".format(f"{str(_try_relative_path(s))}")
+            f"{str(_try_relative_path(s))!r}"
             for s in reversed(self.source_trail)
             if s is not None
         )
--- a/Lib/fontmake/errors.py
+++ b/Lib/fontmake/errors.py
@@ -26,7 +26,7 @@ class FontmakeError(Exception):
 
     def __str__(self):
         trail = " -> ".join(
-            f"'{str(_try_relative_path(s))}'"
+            "{!r}".format(f"{str(_try_relative_path(s))}")
             for s in reversed(self.source_trail)
             if s is not None
         )
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -225,11 +225,10 @@ class Instantiator:
         # point of reference for all data.
         default_font = designspace.default.font
         non_default_layer_name = designspace.default.layerName
-        default_source_uses_non_default_layer = non_default_layer_name is not None
 
         glyph_names: Set[str] = set(default_font.keys())
 
-        if default_source_uses_non_default_layer:
+        if non_default_layer_name is not None:
             try:
                 layer = default_font.layers[non_default_layer_name]
             except KeyError as e:
@@ -240,20 +239,27 @@ class Instantiator:
             layer = default_font.layers[non_default_layer_name]
             glyph_names = layer.keys()
             logger.info(f"Building from layer {layer.name}")
-        else:
-            for source in designspace.sources:
-                other_names = set(source.font.keys())
-                diff_names = other_names - glyph_names
-                if diff_names:
-                    logger.warning(
-                        "The source %s (%s) contains glyphs that are missing from the "
-                        "default source, which will be ignored: %s. If this is unintended, "
-                        "check that these glyphs have the exact same name as the "
-                        "corresponding glyphs in the default source.",
-                        source.name,
-                        source.filename,
-                        ", ".join(sorted(diff_names)),
-                    )
+
+        for source in designspace.sources:
+            other_names = set(source.font.keys())
+            diff_names = other_names - glyph_names
+            if diff_names:
+                max_diff_glyphs = 10
+                logger.warning(
+                    "The source %s (%s)%s contains glyphs that are missing from the "
+                    "default source, which will be ignored: %s%s; if this is unintended, "
+                    "check that these glyphs have the exact same name as the "
+                    "corresponding glyphs in the default source.",
+                    source.name,
+                    source.filename,
+                    f" [layer: {source.layerName}]"
+                    if non_default_layer_name is not None
+                    else "",
+                    ", ".join(sorted(diff_names)[0:max_diff_glyphs]),
+                    f"... ({len(diff_names)} total)"
+                    if len(diff_names) > max_diff_glyphs
+                    else "",
+                )
 
         # Construct Variators
         axis_bounds: AxisBounds = {}  # Design space!
@@ -534,9 +540,8 @@ def collect_info_masters(
     locations_and_masters = []
 
     for source in designspace.sources:
-        if source.layerName is not None:
-            if source is not designspace.default:
-                continue  # No font info in source layers.
+        if source.layerName is not None and source is not designspace.default:
+            continue  # No font info in non-default source layers.
 
         normalized_location = varLib.models.normalizeLocation(
             source.location, axis_bounds
@@ -560,9 +565,8 @@ def collect_kerning_masters(
     locations_and_masters = []
 
     for source in designspace.sources:
-        if source.layerName is not None:
-            if source is not designspace.default:
-                continue  # No kerning in source layers.
+        if source.layerName is not None and source is not designspace.default:
+            continue  # No kerning in non-default source layers.
 
         # If a source has groups, they should match the default's.
         if source.font.groups and source.font.groups != groups:
--- a/Lib/fontmake/instantiator.py
+++ b/Lib/fontmake/instantiator.py
@@ -147,8 +147,6 @@ UFO_INFO_ATTRIBUTES_TO_COPY_TO_INSTANCES = {
     "woffMinorVersion",
 }
 
-DEFAULT_LAYER_NAME = "foreground"
-
 
 # Custom exception for this module
 class InstantiatorError(Exception):
@@ -227,18 +225,21 @@ class Instantiator:
         # point of reference for all data.
         default_font = designspace.default.font
         non_default_layer_name = designspace.default.layerName
-        non_default_layer_flag = (
-            non_default_layer_name is not None
-            and non_default_layer_name != DEFAULT_LAYER_NAME
-        )
+        default_source_uses_non_default_layer = non_default_layer_name is not None
 
         glyph_names: Set[str] = set(default_font.keys())
 
-        if non_default_layer_flag:
-            if non_default_layer_name in default_font.layers:
+        if default_source_uses_non_default_layer:
+            try:
                 layer = default_font.layers[non_default_layer_name]
-                glyph_names = layer.keys()
-                logger.info(f"Building from layer {layer.name}")
+            except KeyError as e:
+                raise InstantiatorError(
+                    f"Layer {non_default_layer_name!r} not found "
+                    f"in {designspace.default.filename}"
+                ) from e
+            layer = default_font.layers[non_default_layer_name]
+            glyph_names = layer.keys()
+            logger.info(f"Building from layer {layer.name}")
         else:
             for source in designspace.sources:
                 other_names = set(source.font.keys())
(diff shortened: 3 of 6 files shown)


Lib/fontmake/__main__.py
```diff
@@ -137,7 +137,7 @@ def parse_mutually_exclusive_inputs(parser, args):
         ):
             ufo_paths.append(filename)
         else:
-            parser.error(f"Unknown input file extension: '{filename}'")
```
anthrotype, line 140: don't need to use `.format()`, you can keep the f-string and add `!r` to get the repr, e.g. `f"{filename!r}"`

Lib/fontmake/errors.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, msg, source_file):
 
     def __str__(self):
         trail = " -> ".join(
-            f"'{str(_try_relative_path(s))}'"
+            "{!r}".format(f"{str(_try_relative_path(s))}")
```
anthrotype, line 29: same as above.. Did the linter suggest you to apply this change? If so, surprised it didn't suggest to keep f-strings (which are more efficient)

Lib/fontmake/instantiator.py
```diff
@@ -147,6 +147,8 @@
     "woffMinorVersion",
 }
 
+DEFAULT_LAYER_NAME = "foreground"
```
anthrotype, line 150: this is not standard, it might be what RoboFont uses maybe but it's not in the UFO spec.

verbosus:
The build failure seems to be related to some codecov “owner” argument I don’t know much about.

Lib/fontmake/instantiator.py
```diff
@@ -224,21 +226,33 @@ def from_designspace(
         # because the math behind varLib and MutatorMath uses the default font as the
         # point of reference for all data.
         default_font = designspace.default.font
+        non_default_layer_name = designspace.default.layerName
+        non_default_layer_flag = (
+            non_default_layer_name is not None
+            and non_default_layer_name != DEFAULT_LAYER_NAME
```
anthrotype, line 232: you should remove this `and non_default_layer_name != DEFAULT_LAYER_NAME`, it's not necessary. If the source.layerName is not None, then we can simply use it at face value, whatever that was. "foreground" or "public.default" aren't special (the latter is simply the name of an un-named default layer). What makes a layer the default is the fact that its directory is named 'glyphs' without any suffix.

Lib/fontmake/instantiator.py
```diff
@@ -224,21 +226,33 @@ def from_designspace(
...
+            if non_default_layer_name in default_font.layers:
+                layer = default_font.layers[non_default_layer_name]
+                glyph_names = layer.keys()
+                logger.info(f"Building from layer {layer.name}")
+        else:
+            for source in designspace.sources:
+                other_names = set(source.font.keys())
+                diff_names = other_names - glyph_names
+                if diff_names:
+                    logger.warning(
```
anthrotype, line 247: I think this check that there aren't different glyphs in non-default masters, we would like to do regardless of whether the default master is taken from a UFO's default layer or a non-default layer, so we should move the whole for block one indentation level up

Lib/fontmake/instantiator.py
```diff
@@ -224,21 +226,33 @@ def from_designspace(
         # because the math behind varLib and MutatorMath uses the default font as the
         # point of reference for all data.
         default_font = designspace.default.font
+        non_default_layer_name = designspace.default.layerName
+        non_default_layer_flag = (
```
anthrotype, line 230: instead of `non_default_layer_flag`, how about `default_source_uses_non_default_layer`.. a bit long maybe but it kind of reads better when you see it in e.g. `if default_source_uses_non_default_layer:`

Lib/fontmake/instantiator.py
```diff
@@ -224,21 +226,33 @@ def from_designspace(
...
-                    "The source %s (%s) contains glyphs that are missing from the "
-                    "default source, which will be ignored: %s. If this is unintended, "
-                    "check that these glyphs have the exact same name as the "
-                    "corresponding glyphs in the default source.",
-                    source.name,
-                    source.filename,
-                    ", ".join(sorted(diff_names)),
-                )
+        if non_default_layer_flag:
+            if non_default_layer_name in default_font.layers:
```
anthrotype, line 238: you are not handling the `else:` here. We don't want to silently pass if the specified surce.layerName actually is not in the font. We want to explode! 
Do:

```python
try:
    layer = default_font.layers[non_default_layer_name]
except KeyError as e:
    raise InstantiatorError(f"Layer {non_default_layer_name!r} not found in {source.filename}") from e
```

Lib/fontmake/instantiator.py
```diff
@@ -517,9 +531,11 @@ def collect_info_masters(
 ) -> List[Tuple[Location, FontMathObject]]:
     """Return master Info objects wrapped by MathInfo."""
     locations_and_masters = []
+
     for source in designspace.sources:
-        if source.layerName is not None:
-            continue  # No font info in source layers.
+        if source.layerName is not None and source.layerName != DEFAULT_LAYER_NAME:
+            if source != designspace.default:
```
anthrotype, line 537: don't use the name to check if a layer is the default, also it's not relevant here what the name of a source layer is. Here we simply want to skip any non-default sources, so the latter check is sufficient (better to use `is not` in this case): `if source is not designspace.default`.

Lib/fontmake/instantiator.py
```diff
@@ -541,9 +557,11 @@ def collect_kerning_masters(
     groups = designspace.default.font.groups
 
     locations_and_masters = []
+
     for source in designspace.sources:
-        if source.layerName is not None:
-            continue  # No kerning in source layers.
+        if source.layerName is not None and source.layerName != DEFAULT_LAYER_NAME:
+            if source != designspace.default:
```
anthrotype, line 563: same thing here as for the font info above; ignore layerName, just continue `if source is not designspace.default`

anthrotype (commented):
hi Antonio! thanks, I left some comments

Lib/fontmake/instantiator.py
```diff
@@ -396,7 +396,7 @@ def generate_instance(
                 # whatever reason (usually outline incompatibility)...
                 if glyph_name not in self.skip_export_glyphs:
                     raise InstantiatorError(
-                        f"Failed to generate instance of glyph '{glyph_name}': "
+                        f"Failed to generate instance of glyph {glyph_name}: "
```
anthrotype, line 399: here we lost the quotes, you meant to do this

```diff
+ f"Failed to generate instance of glyph {glyph_name}: "
- f"Failed to generate instance of glyph {glyph_name!r}: "
```

Lib/fontmake/instantiator.py
```diff
@@ -293,7 +293,7 @@ def from_designspace(
                 glyph_mutators[glyph_name] = Variator.from_masters(items, axis_order)
             except varLib.errors.VarLibError as e:
                 raise InstantiatorError(
-                    f"Cannot set up glyph '{glyph_name}' for interpolation: {e}'"
+                    f"Cannot set up glyph {glyph_name} for interpolation: {e}'"
```
anthrotype, line 296: we lost the quotes around glyph_name. Also, looks like we left a trailing `'` at the end for some reasons, probably a typo

```diff
+ f"Cannot set up glyph {glyph_name} for interpolation: {e}'"
- f"Cannot set up glyph {glyph_name!r} for interpolation: {e}"
```

anthrotype:
don't worry about codecov, it's flaky at best.. i'm going to make the PR pass even when it fails to upload (which is 50% of the time lately)

Lib/fontmake/__main__.py
verbosus: I see thanks, I followed what flake8 said without much thinking, and did not know !r was available as a modifier to  another {variable}. Fixed, pushing shortly.

Lib/fontmake/instantiator.py
verbosus: OK, fixed.

Lib/fontmake/instantiator.py
verbosus: OK fixed.

Lib/fontmake/instantiator.py
verbosus: OK fixed.

Lib/fontmake/instantiator.py
verbosus: OK fixed.

Lib/fontmake/instantiator.py
verbosus: OK.

Lib/fontmake/instantiator.py
verbosus: Yeah, it’s the RoboFont default. Removed.

Lib/fontmake/instantiator.py
verbosus: Yeah, same as above. Fixed.

verbosus:
Thanks Cosimo: I believe I fixed all your comments, except the one about reporting the non-existing glyphs in non-default masters. I’d rather have it behave like it does right now: it’s a corner case anyway so the main path isn’t affected, and for the way I’m using it I much prefer it to silently ignore the missing glyphs rather than reporting them. Codecov still isn’t happy, but I don’t think there is anything I can do to help there.

Lib/fontmake/instantiator.py
```diff
@@ -224,21 +224,36 @@ def from_designspace(
...
-                logger.warning(
-                    "The source %s (%s) contains glyphs that are missing from the "
-                    "default source, which will be ignored: %s. If this is unintended, "
-                    "check that these glyphs have the exact same name as the "
-                    "corresponding glyphs in the default source.",
-                    source.name,
-                    source.filename,
-                    ", ".join(sorted(diff_names)),
-                )
+        if default_source_uses_non_default_layer:
```
anthrotype, line 232: maybe we can remove this local variable `default_source_uses_non_default_layer` since it's only used once and simply say

```python
if non_default_layer_name is not None:
```

Lib/fontmake/instantiator.py
anthrotype: > This makes the build extremely noisy

I don't understand how this should be any different whether we use the default layer or a non-default one. If you think it's too noisy, we can print the first 10 or so and truncate with ellipsis.

anthrotype: the effect of having additional glyphs in the non-default sources that are not in the default one is the same, whether or not the default source is taken from the default layer or a non-default layer: those glyphs will be ignored. I think it still warrants a warning.

Lib/fontmake/instantiator.py
```diff
@@ -517,9 +532,11 @@ def collect_info_masters(
 ) -> List[Tuple[Location, FontMathObject]]:
     """Return master Info objects wrapped by MathInfo."""
     locations_and_masters = []
+
     for source in designspace.sources:
         if source.layerName is not None:
```
anthrotype, line 537: I think we want to get rid of this `if source.layerName is not None:` altogether, and only rely on the check `if source is not designspace.default:` in order to decide whether to consider this source's info or kerning or whether to skip it when it's not the default

anthrotype: oh sorry sorry! I misunderstood.. this is gathering sources from all masters that contribute to info/kerning and we were previously skipping the 'sparse' sources.. let me think

Lib/fontmake/instantiator.py
verbosus: Agreed, I’ve removed it. Thanks!

Lib/fontmake/instantiator.py
anthrotype: ok yes you're right, we need to check both layerName and whether source is not default. Please put the two if conditions on the same line using `and` operator

Lib/fontmake/instantiator.py
```diff
@@ -517,9 +532,11 @@ def collect_info_masters(
...
-            continue  # No font info in source layers.
+            if source is not designspace.default:
+                continue  # No font info in source layers.
```
anthrotype, line 539:
```suggestion
                continue  # No font info in non-default source layers.
```

Lib/fontmake/instantiator.py
verbosus: I was just about to finish writing a comment to the same effect. Fixed, thanks!

Lib/fontmake/instantiator.py
verbosus: Fixed.

Lib/fontmake/instantiator.py
verbosus: The above logic is what I just checked in for review.

anthrotype:
this is looking good now, thanks!
Maybe we should change the PR title to better reflect the changes that you made. Only the fontmake.instantiator (which generates static instance UFOs from a DS) has been modified, whereas the PR title suggests building in general (not just static interpolated instances) from non-default layer source isn't supported..

by the way, did it work before this PR to build a VF from the same DS with such non-default layer set as default source?

anthrotype:
I'm also curious, why not simply mark the desired non-default layer that you wish to use as the default DS source as _the default_ layer in the font editor or via script?

verbosus:
In some cases it’s handy to have metrically-compatible sources in the same UFO (think: a full outline and a stencil variant, etc.) but on different layers. And yeah, I tried and it did NOT use to work before the PR: fontmake was crashing saying it couldn’t find a base master.

anthrotype:
> I tried and it did NOT use to work before the PR: fontmake was crashing saying it couldn’t find a base master.

was that trying to build interpolated static instance font (via instantiator) or building a VF (which does not go through instantiator)?

I'm trying to understand if this PR fixes other things, or if other parts of the pipeline need similar fixes to adjust expectation that a base source may not necessarily come from a default layer

verbosus:
Sorry, I misunderstood your previous question. I only tried building static instances so far. I just checked and building a VF fails either with or without this PR. The failure is the same: `fontmake: Error: In 'sources/testfamily.designspace': Generating fonts from Designspace failed: "'name' table not found"`. Unless it raises any flags for obvious known issues, I guess it needs some deeper investigation.

anthrotype:
@verbosus can you try running fontmake with `--verbose DEBUG` option so it displays the full exception traceback, so you can see where exactly that error is raised?

anthrotype:
@verbosus please file an issue about that error when you find out more details, and thanks again for the PR

verbosus:
Thanks for merging and sure, I just filed #983 to cover the rest of the work on the variable path.

PR #1017, opened 2023-07-20
Also open JSON UFOs

madig:


madig:
Hm. Maybe we want a (hard|optional) dep on `ufoLib2[json,msgpack]`?

Lib/fontmake/font_project.py
```diff
@@ -128,6 +129,11 @@ def __init__(
 
     def open_ufo(self, path):
         try:
+            path = Path(path)
+            if path.suffix == ".json":
+                with open(path, "rb") as f:
+                    # pylint: disable=no-member
+                    return ufoLib2.Font.json_load(f)  # type: ignore
```
anthrotype, line 136: @madig do we actually need these pylint/type:ignore comments? We aren't checking them, are we?
If you think it's important to have them, I can add to the other call site below where we do `font.json_dump()` which I think will suffer from the same issues with linters

madig: It only matters if you use pylint/mypy to check it.

anthrotype: I just added them https://github.com/googlefonts/fontmake/pull/1017/commits/920f55eae8634f0f0e379c93e09bc3acbe23d65a

anthrotype:
> Maybe we want a (hard|optional) dep on ufoLib2[json]?

in the end, I added an optional 'json' to `extras_require` which fetches ufoLib2[json]

PR #1021, opened 2023-07-31
add --ttf-curves to support building glyf v1 with cubics

anthrotype:
The new option allows to pass 'mixed' (to selectively convert cubics to quad only when more economical), 'keep-cubic' / 'keep-quad' (to skip conversion altogether and keep the desired input curve format). 
The default value 'cu2qu' keeps the current behavior unchanged (i.e. all cubics converted to quad and glyf v0).

Requires ufo2ft >= 2.33.2


anthrotype:
I added a note to both --help and docstrings that support for glyf v1 is experimental and spec proposal (https://github.com/harfbuzz/boring-expansion-spec/blob/main/glyf1.md) not finalized, and as such not widely implemented.
It's also impossible to accidentally produce a glyf v1 table, because even if cu2qu is skipped with --keep-quad, ufo2ft will return an error if any of the glyphs do contain cubics but the head.glyphDataFormat = 0

PR #1027, opened 2023-08-07
Added Names of Authors and contributors in setup.py

ChibuezeOnejeme:
I was working with Pydigger API  project that tracks different projects from PYPI that show various stats including incomplete packaging info, fontmake popped up and I decided to update the information.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
     long_description=long_description,
     author="James Godfrey-kittle",
     maintainer=" Cosimo Lupo",
-   
+    maintainer_email="cosimo@anthrotype.com",
     long_description_content_type="text/markdown",
     url="https://github.com/googlei18n/fontmake",
     license="Apache Software License 2.0",
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup(
     ),
     long_description=long_description,
     author="James Godfrey-kittle",
-    maintainer=" Cosimo Lupo",
+    maintainer="Cosimo Lupo",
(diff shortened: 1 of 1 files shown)


anthrotype:
thanks, but I don't think we want to need to mention all the contributors in the author field, it's sufficient we write the original `author`, James Godfrey-Kittle, and perhaps myself "Cosimo Lupo" as the `maintainer`.

setup.py
```diff
@@ -86,36 +86,9 @@
...
-        "Thomas Rettig,"
-        "Chris Simpkins,"
-        "Adrien Tetar"
-    ),
-
-
-    
+    author="James Godfrey-kittle",
+    maintainer=" Cosimo Lupo",
+
```
anthrotype, line 91: please don't leave the empty line or black auto-formatter is going to bark at you.
Also you can add 

```
maintainer_email="cosimo@anthrotype.com",
```

anthrotype (approved):

setup.py
```diff
@@ -86,6 +86,9 @@
         "Compile fonts from sources (UFO, Glyphs) to binary (OpenType, TrueType)."
     ),
     long_description=long_description,
+    author="James Godfrey-kittle",
+    maintainer=" Cosimo Lupo",
```
anthrotype, line 90:
```suggestion
    maintainer="Cosimo Lupo",
```

anthrotype (changes requested):

anthrotype (approved):

PR #1043, opened 2023-09-26
Fix glyph swap

yanone:
Fixing duplication of rule-based glyph swaps


yanone:
About the build fail:

Please help me make sense of this: in `test_rules_are_applied_deterministically()` it appears that rules _should_ be applied several times, as the test checks to compare against outlines of `"Q.ss01.alt"`. 

I added a warning to the main code that prints:
`WARNING  root:instantiator.py:174 Duplicate rule for ('Q', 'Q.ss01') in location {'Weight': 900.0, 'Style': 0.5}: ('Q', 'Q.alt')`, meaning that `('Q', 'Q.alt')` was already applied (= in the list), and now `('Q', 'Q.ss01')` will be rejected for duplication.

If we're on the correct trajectory here, I think I should indeed check for the whole tuple to already exist, not just for `oldName`.

anthrotype:
@belluzj added that test in https://github.com/googlefonts/fontmake/pull/689 when he fixed some bug about processing rules, maybe he could also take a quick look?

PR #1081, opened 2024-03-25
replace fontmake.instantiator with ufo2ft.instantiator

anthrotype:
fontmake.instantiator module now lives in ufo2ft.instantiator (https://github.com/googlefonts/ufo2ft/pull/825); the current fontmake.instantiator module is kept as a stub that re-exports the symbols from the new ufo2ft.instantiator for backward compatibility. I also kept the test_instantiator.py to make sure it continues to work unchanged.

The attrs dependency is no longer needed, since ufo2ft.instantiator uses dataclasses, so I removed it from fontmake's direct dependencies (it's still used indirectly in ufoLib2, which fontmake depends on).


anthrotype:
test fails because it needs https://github.com/googlefonts/glyphsLib/pull/992, need to first do a glyphsLib release..

PR #1090, opened 2024-05-09
`CompatibilityChecker`: check all layers within designspace sources

RickyDaMa:
The problem is that the compatibility checker iterates over every source but does not consider whether they are sparse layers

This means that it duplicates work and misrepresents how many sources it's checking (given that it counts every layer, but only checks the default layer for each UFO)

This PR changes the reporting to be accurate (and improves phrasing / word choice to aid understanding), and **extends the `CompatibilityChecker`'s scope to include the previously unchecked non-default layers**

There's definitely cases where more checking is more betterer, however it'll creating additional compatibility check failures (even in the test data used within this repository). Before continuing this PR with tests etc., we want to make sure that the checks aren't now overstepping from correctness into things that should be warnings (e.g. are mismatching anchors in sparse layers worth failing the build over?). We'll add & update tests once we've discussed & decided

Aside: https://matklad.github.io/2022/10/24/actions-permissions.html

### DIFF
--- a/Lib/fontmake/compatibility.py
+++ b/Lib/fontmake/compatibility.py
@@ -100,15 +100,15 @@ class CompatibilityChecker:
         for obj, (font, layer) in zip(objs, self.current_layers):
             values.setdefault(func(obj), []).append(self._name_for(font, layer))
         if len(values) < 2:
-            logger.debug(f"All fonts had same {what} in {context}")
+            logger.debug(f"All sources had same {what} in {context}")
             return True
-        report = f"\nFonts had differing {what} in {context}:\n"
+        report = f"\nSources had differing {what} in {context}:\n"
         debug_enabled = logger.isEnabledFor(logging.DEBUG)
         for value, source_names in values.items():
             if debug_enabled or len(source_names) <= 6:
                 key = ", ".join(source_names)
             else:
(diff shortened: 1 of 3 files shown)


anthrotype:
> Anchors - Discussed above, we need to stop checking these

@RickyDaMa you should first actually check that it works as intended in ufo2ft and doesn't crash. this remindes me https://github.com/googlefonts/ufo2ft/pull/840

anthrotype:
> GPOS context for anchors

what's that?

anthrotype:
ah it's the glyphsLib's ContextualMarkFeatureWriter 

https://github.com/googlefonts/glyphsLib/blob/cedaacdc2bd17e879c3332b30e567ce5bafa0f37/Lib/glyphsLib/featureWriters/markFeatureWriter.py#L116

well I suppose whatever treatment you apply for anchors in general you should also apply for their GPOS context

anthrotype:
>> Anchors - Discussed above, we need to stop checking these

> RickyDaMa you should first actually check that it works as intended in ufo2ft and doesn't crash.

@RickyDaMa actually this works as of now thanks to @khaledhosny https://github.com/googlefonts/ufo2ft/pull/842

PR #1154, opened 2025-08-27
Pass auto_use_my_metrics=False to ufo2ft for variable fonts

anthrotype:
This is fix/prevent the issue described in https://github.com/fonttools/fonttools/pull/3912

Also see the discussion in https://github.com/googlefonts/fontc/issues/1614

The PR is intended to [match fontc's default behavior](https://github.com/googlefonts/fontc/pull/1615) whereby the USE_MY_METRICS glyf component flag is only automatically set for static font builds, and no longer also for variable font builds, where this can be problematic when composite/component glyph metrics are inconsistent across masters and the HVAR table (normally preferred over glyf+gvar phantom points) gets dropped.


anthrotype:
fontmake already had a `--no-auto-use-my-metrics` CLI flag to disable this. The PR adds a positive `--auto-use-my-metrics` flag as well, to manually opt-in the ufo2ft feature.

By default, fontmake will continue to pass `auto_use_my_metrics=True` when building static fonts (either from `-u` single ufos, or from interpolated instances with the `-i` flag).
But it will pass `auto_use_my_metrics=False` when building interpolatable master TTFs or variable fonts.

cmyr (approved):
looks reasonable!

PR #1168, opened 2026-05-26
Generate preliminary openTypeCategories from GlyphData.xml for DS/UFOs

anthrotype:
When building DS+UFO sources that lack explicit `public.openTypeCategories`, fontmake now generates preliminary categories for marks and ligatures using glyphsLib's GlyphData.xml and passes them to ufo2ft via the new
`preliminaryOpenTypeCategories` kwarg from https://github.com/googlefonts/ufo2ft/pull/985.

ufo2ft finalizes them after anchor propagation, pruning anchorless ligatures and inferring bases from attaching anchors.

This closes the gap between .glyphs and DS+UFO sources: .glyphs sources get categories via glyphsLib's export, but DS+UFO sources authored directly had no way to get them automatically. Without categories, ufo2ft's anchor propagation would fall back to fallible heuristics, leading to incorrect ligature anchor numbering and GDEF differences vs fontc.

`_preliminary_categories_for_font(font)` iterates over all glyphs and classifies them using `glyphsLib.glyphdata.get_glyph()`, respecting per-glyph category/subCategory overrides from the glyph's lib (matching Glyphs.app behavior). Marks are glyphs with category "Mark" and subCategory "Nonspacing" or "Spacing Combining"; ligatures are glyphs with subCategory "Ligature".

The categories are "preliminary" because they're based on GlyphData.xml alone, before anchor propagation. ufo2ft refines them: marks are kept as-is, ligatures are kept only if they gained attaching anchors after propagation, and bases are inferred for remaining glyphs with attaching anchors. This classification logic matches glyphsLib's [`_build_public_opentype_categories`](https://github.com/googlefonts/glyphsLib/blob/75c07d42/Lib/glyphsLib/builder/features.py#L213-L272).

Sources that already have explicit `public.openTypeCategories` (in the DS lib or font lib) are left untouched.

Depends on googlefonts/ufo2ft#985

### DIFF
--- a/Lib/fontmake/font_project.py
+++ b/Lib/fontmake/font_project.py
@@ -114,6 +114,92 @@ class CurveConversion(enum.Enum):
         )
 
 
+OPENTYPE_CATEGORIES_KEY = PUBLIC_PREFIX + "openTypeCategories"
+
+try:
+    from dataclasses import fields as _dc_fields
+
+    from ufo2ft._compilers.baseCompiler import BaseCompiler as _BC
+
+    _HAS_PRELIMINARY_CATEGORIES = any(
+        f.name == "preliminaryOpenTypeCategories" for f in _dc_fields(_BC)
+    )
+except Exception:
+    _HAS_PRELIMINARY_CATEGORIES = False
+
+
+def _preliminary_categories_for_font(font):
+    """Generate preliminary openTypeCategories from GlyphData.xml for a font.
+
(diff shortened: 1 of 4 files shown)


anthrotype:
Note I have NOT bumped the minimum ufo2ft. If for whatever reasons you prefer to pin ufo2ft to the previous version with the old `propagateAnchors` filter, you can still do that.

PR #1173, opened 2026-09-10
[compatibility] Normalize contour points before comparing

### LINKED ISSUE #875: ERROR:fontmake.compatibility – compatibility checker is maybe overzealous picky about start points
The compatibility checker seems to block glyphs that will interpolate without issue.

I have at least three projects which will build with FontMake `3.0.0`, but are blocked in `3.1.0`. This includes Recursive, an OFL project, though it is slightly difficult to make a direct reproduction repo because its build is more complex than just a FontMake command. But, if it would be necessary, I could probably make an abbreviated OFL reproduction repo. It’s possible that all of these are blocked because they have been drawn in RoboFont, and I’m not sure whether fonts drawn in e.g. Glyphs or FontLab would have the same issue. However, there are a good number of OFL fonts commissioned by Google that were drawn in RoboFont (Fraunces, Georama, Roboto Serif, Spectral...), so I suspect this issue probably has a fairly wide impact.

The main thing the compatibility checker is getting hung up on are `differing point type` errors, which basically come down to some glyphs having oncurve vs offcurve start points, though sharing the same start segment. For example, here is an outline that has an offcurve point as point index `0`, whereas FontMake expects the start point to be an oncurve:

image

FontMake’s expectation of point `0` being an oncurve – or at least matching exactly between all sources – seems sensible enough. However, it _will_ build these glyphs without issue in FontMake version `3.0.0`. 

As a concrete example, the compatibility checker in `v3.1.0` will block the build for `g.text`, `nine`, `oe`, and more in my Name Sans project:

```
ERROR:fontmake.compatibility:Fonts had differing point type in glyph g.text, contour 1, point 0:
ERROR:fontmake.compatibility: * Name Sans Text Hairline Italic, Name Sans Text Ultra Italic had curve
ERROR:fontmake.compatibility: * Name Sans Text Bold Italic had None
ERROR:fontmake.compatibility:Fonts had differing point type in glyph g.text, contour 1, point 2:
ERROR:fontmake.compatibility: * Name Sans Text Hairline Italic, Name Sans Text Ultra Italic had None
ERROR:fontmake.compatibility: * Name Sans Text Bold Italic had curve
ERROR:fontmake.compatibility:Fonts had differing point type in glyph g.text, contour 1, point 3:
ERROR:fontmake.compatibility: * Name Sans Text Hairline Italic, Name Sans Text Ultra Italic had curve
ERROR:fontmake.compatibility: * Name Sans Text Bold Italic had None
# etc etc
```

However, in `v3.0.0`, these glyphs build into a VF with no visible interpolation issues (at least that I can see in FontGoggles, or that I have seen or heard reported anywhere else):

https://user-images.githubusercontent.com/45946693/163728494-ec948d4d-a09a-4b99-ab9b-719e593efabc.mp4

Of course, there may be good reasons to make these start points exactly match. It certainly seems logical. But, were there actual, documented issues caused by glyphs that had start points that were offcurve in some sources and oncurve in others? Or, could the compatibility checker be made slightly more permissive?

Thanks, as always, to all the FontMake maintainers for their work!

cmyr:
Been looking at the various cases on crater where fontmake fails and fontc doesn't, and came across this.

CompatibilityChecker walks masters index-by-index over the raw UFO point list. This is not the same logic used by the point pen in fontTools, which rotates the point list so that it ends on the first on-curve point (see https://github.com/fonttools/fonttools/blob/82cd560fd9/Lib/fontTools/pens/pointPen.py#L186-L188 ).

This meant that the CompatibilityChecker would reject curves that were actually compatable, when two masters had serialized the same point list with a different start point.

This makes CompatibilityChecker match the rotation logic of the point pen, which means fontmake can now compile a number of fonts that it would previously fail on, such as Exo, Raleway-Italic, and RobotoMono-Italic.

Note that for a contour that really is incompatible, the reported point index is now in normalized rather than glif order.

- fixes #875


anthrotype:
we should not apply this rotation unconditionally.

cu2qu rewrites the glyph through the segment pen (so it actually rotates) only when a cubic was converted or `reverse_direction` is True. When fontmake passes `--keep-direction` or `--ttf-curves keep-quad` (or keep-cubic) the TTGlyphPointPen keeps the raw contour points' order, so a start-point mismatch is a real interpolatable incompatibility.

varLib then does not error in this case but simply warns "glyph has incompatible masters; skipping" and the VF ships with empty gvar glyph with no variations. The keep-direction/keep-quad exists exactly so that point indices stay put (e.g. for manual VTT hinting workflows where that matters).

You could add a `normalize_start_points` kwarg (default True), and at `run_from_designspace` pass True only when we aren't emitting TTF, or when `ttf_curves.convertCubics` and `reverse_direction` are _both_ on.

`--keep-direction` with the default `--ttf-curves cu2qu` would still flag today's #875 false positives for the glyphs that do contain cubic curves (cu2qu does rotate those); but I don't think a per-glyph "modified iff any cubic" would be worth it.

> the reported point index is now in normalized rather than glif order

maybe say "point N (counting from the first on-curve point)" so it isn't read as a glif index

anthrotype (approved):
LGTM

